PR #65, opened 2020-10-27
individual biosamples, overview, and diseases

brouillette:
- updates


src/components/explorer/IndividualBiosamples.js
```diff
@@ -51,7 +51,7 @@ const IndividualBiosamples = ({individual}) =>
            pagination={{pageSize: 25}}
            columns={BIOSAMPLE_COLUMNS}
            rowKey="id"
-           dataSource={(individual || {}).biosamples || []} />;
+           dataSource={(individual || {}).phenopackets.map(p => p.biosamples).flat() || []} />;
```
davidlougheed, line 54: this could be shortened to `.flatMap(...`

src/components/explorer/IndividualDiseases.js
```diff
@@ -0,0 +1,59 @@
...
+    // },
+];
+
+const IndividualDiseases = ({individual}) =>
+    <Table bordered
+           size="middle"
+           pagination={{pageSize: 25}}
+           columns={DISEASE_COLUMNS}
+           rowKey="id"                    // TODO: map all phenopackets instead of just displaying the first
+           dataSource={(individual || {}).phenopackets.map(p => p.diseases).flat() || []} />;
```
davidlougheed, line 51: same here

src/components/explorer/IndividualBiosamples.js
davidlougheed: also, this can trip an error, since the `|| []` 
thing will not kick in if `.map` throws an error
however, i dont think its needed, since these fields should be always present, so the `|| []` at the end could probably just be removed

davidlougheed (changes requested):
some minor things, overall looks good though!

davidlougheed (approved):
see below re-review

src/components/explorer/IndividualDiseases.js
```diff
@@ -0,0 +1,58 @@
+import React from "react";
+
+import "antd/es/table/style/css";
+import {Table} from "antd";
+
+
+import {EM_DASH} from "../../constants";
```
davidlougheed, line 7: lint labels this as unused

src/components/explorer/IndividualDiseases.js
```diff
@@ -0,0 +1,58 @@
...
+import {renderOntologyTerm} from "./ontologies";
```
davidlougheed, line 8: as well as this

davidlougheed (changes requested):

src/components/explorer/IndividualDiseases.js
brouillette: Table is used in rendering, but will remove EM_DASH

src/components/explorer/IndividualDiseases.js
brouillette: No prob

src/components/explorer/IndividualDiseases.js
davidlougheed: yeah i just meant EM_DASH! thank you

davidlougheed (approved):

PR #68, opened 2020-11-09
Merge new re-shuffled dashboard

brouillette:
As per 
https://206.12.92.46/issues/234


davidlougheed:
for the UI update, I think the menu ant design component is only meant to be there once (since otherwise the visual hierarchy is a bit confusing) - would it make sense to use a dropdown from the admin menu for the admin stuff?

brouillette:
@dbujold good point! I'll look into that

brouillette:
closing this PR and will do one larger PR with both the Overview page and the headers taken care of

brouillette:
reopening PR
includes overview container with charts and data summaries

zxenia (approved):
LGTM!

src/constants.js
```diff
@@ -1,4 +1,5 @@
 export const SITE_NAME = "Bento";
+export const ADMIN_NAME = "Admin";
```
davidlougheed, line 2: This probably doesn't need to be "variable-ized" the same way the site name does, since it won't really change

src/modules/auth/actions.js
```diff
@@ -8,14 +8,16 @@ import {
...
-import {fetchProjectsWithDatasetsAndTables} from "../metadata/actions";
+import {fetchExperiments, fetchProjectsWithDatasetsAndTables, fetchVariantTableSummaries} from "../metadata/actions";
 import {fetchNodeInfo} from "../node/actions";
 import {fetchNotifications} from "../notifications/actions";
 import {fetchServicesWithMetadataAndDataTypesAndTablesIfNeeded} from "../services/actions";
 import {fetchRuns} from "../wes/actions";
 import {nop} from "../../utils/misc";
 import {withBasePath} from "../../utils/url";
 
+import {fetchPhenopackets} from "../metadata/actions";
```
davidlougheed, line 19: Put this with other metadata imports

src/modules/metadata/actions.js
```diff
@@ -389,3 +420,29 @@ export const fetchIndividualIfNecessary = individualID => (dispatch, getState) =
     if (individualRecord.isFetching || individualRecord.data) return;  // Don't fetch if already fetching or loaded.
     return dispatch(fetchIndividual(individualID));
 };
+
+
+
+export const fetchPhenopackets = networkAction(() => (dispatch, getState) => ({
+    types: FETCH_PHENOPACKETS,
+    // params: {phenopacketID},
```
davidlougheed, line 428: remove

src/modules/services/reducers.js
```diff
@@ -59,6 +59,7 @@ export const services = (
         logService: null,
         metadataService: null,
         notificationService: null,
+        variantService: null,
```
davidlougheed, line 62: unneeded

src/modules/services/reducers.js
```diff
@@ -89,6 +90,7 @@ export const services = (
                 logService: itemsByArtifact["log-service"] || null,
                 notificationService: itemsByArtifact["notification"] || null,
                 metadataService: itemsByArtifact["metadata"] || null,
+                variantService: itemsByArtifact["variant"] || null,
```
davidlougheed, line 93: unneeded

src/propTypes.js
```diff
@@ -256,6 +257,17 @@ export const phenopacketPropTypesShape = PropTypes.shape({
     updated: PropTypes.string,  // ISO datetime string
 });
 
+export const experimentPropTypesShape = PropTypes.shape({
+    id: PropTypes.string.isRequired,
+    // subject: PropTypes.oneOfType([individualPropTypesShape, PropTypes.string]).isRequired,
```
davidlougheed, line 262: Ideally this can be filled in from what we already know about the endpoint

src/constants.js
brouillette: This actually isn't needed. I removed the admin nav bar

src/modules/metadata/actions.js
brouillette: why? it's used in the overview

src/propTypes.js
brouillette: agreed

src/modules/metadata/actions.js
davidlougheed: specifically the commented out line, sorry

src/propTypes.js
brouillette: I will update this after the first test on signature

PR #77, opened 2020-12-16
Individual variants and genes pages

brouillette:


src/components/explorer/IndividualVariants.js
```diff
@@ -0,0 +1,78 @@
...
+//                 ? (age.hasOwnProperty("age") ? age.age : `Between ${age.start.age} and ${age.end.age}`)
+//                 : EM_DASH;
+//         },
+//     },
+// ];
+
+const IndividualVariants = ({individual}) =>
+{
+    const biosamples = (individual || {}).phenopackets.flatMap(p => p.biosamples);
+    const variantsFlat = biosamples.map(b=>b.variants).flatMap(v => v.map(_v => _v.hgvsAllele));
```
zxenia, line 14: The `hgvsAllele` is hardcoded here but it can be one of four options because it uses the dynamic schema here.
For now, it's ok but in the future, it should be changed otherwise it won't be able to display other types.

zxenia (approved):

src/components/explorer/IndividualVariants.js
zxenia: instead of hardcoding you should get the value of {"allele_type"}

PR #96, opened 2021-02-22
Fixed pointer events in table while searching

sebastianballesteros:
Fixed David's comments regarding the Explorer search table: "When I submit a new search, the table is greyed out, but the links to patient records and the bottom right pagination are still clickable. They shouldn't be usable once a new search has been launched."


brouillette (approved):
nice!

brouillette:
what's up with the linter @sebastianballesteros ?

sebastianballesteros:
> what's up with the linter @sebastianballesteros ?



> what's up with the linter @sebastianballesteros ?

@brouillette A problem with the length of the line. I will fix it.

PR #98, opened 2021-04-01
Fix for docker support + another linting rule

davidlougheed:
Merge this after new versions of bento_wes, proxy, etc are deployed to instances.


davidlougheed:
replaced by https://github.com/bento-platform/bento_web/pull/99

PR #104, opened 2021-05-10
pie chart fixes

gsfk:
Two small Overview pie chart fixes:

1: don't show labels for empty categories 

2: eliminate possibility of adjacent pie segments with the same colour


gsfk:
closing, will integrate into Overview fixes

PR #113, opened 2021-07-08
Front-end changes for Katsu case-insensitive search

gsfk:
update search equality operator to allow for Katsu case-insensitive search. Needs testing.

### DIFF
--- a/src/utils/search.js
+++ b/src/utils/search.js
@@ -1,6 +1,6 @@
 import {simpleDeepCopy} from "./misc";
 
-export const OP_EQUALS = "eq";
+export const OP_EQUALS = "ico";
 export const OP_LESS_THAN = "lt";
 export const OP_LESS_THAN_OR_EQUAL = "le";
 export const OP_GREATER_THAN = "gt";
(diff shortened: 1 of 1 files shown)


zxenia (approved):
LGTM!

davidlougheed:
This is incorrect since ico is contains not equals

davidlougheed:
Instead, add new operators and change which one is the default if you want to make it easier to search

davidlougheed:
To go a bit more into detail, you should never be changing those constants except if an operator is renamed in bento_lib; they are just supposed to be JS representations of the operator strings themselves (eq = OPERATOR_EQUALS, etc)

I suggest taking a closer look at the search schemas coming from the different data services and seeing how the front end uses some of these fields to build the search interface.

zxenia:
@davidlougheed just to clarify a little bit here, are those changes mainly should be done on the front-end (bento_web) or should I change those operators in katsu or in bento_lib?

gsfk:
Draft of a more appropriate fix. Needs a few more changes if text-insensitive search is meant as default. Needs testing.

brouillette:
LGTM!

PR #116, opened 2021-07-22
[WIP] Experiments search support

davidlougheed:


src/modules/explorer/reducers.js
```diff
@@ -33,13 +43,17 @@ const tableSearchResults = (searchResults) => {
...
                 phenotypic_features: {},
-                experiments: [],  // TODO
+                experiments: {},  // Filled from all the experiments conducted on the biosamples
             };
         }
 
-        (p.biosamples || []).forEach(b => tableResultSet[individualID].biosamples[b.id] = b);
-        (p.diseases || []).forEach(d => tableResultSet[individualID].diseases[d.id] = d);
-        (p.phenotypic_features || []).forEach(pf => tableResultSet[individualID].phenotypic_features[pf.type.id] = pf);
+        (p.biosamples ?? []).forEach(b => {
```
brouillette, line 50: Should this not be `p.biosamples || []` instead `of ??` ?

davidlougheed: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator

in this case they almost do the same thing, but since we're just using it for undefined/missing keys i think `??` makes more sense going forward because it is more explicit in its semantic meaning vs logical or

brouillette: I see!
:+1:

brouillette:
LGTM!

PR #120, opened 2021-07-30
fix age in Individual Overview

gsfk:
Very quick fix for age in Individual overview


src/components/explorer/IndividualOverview.js
```diff
@@ -11,7 +11,7 @@ const IndividualOverview = ({individual}) => individual ?
     <Descriptions layout="vertical" bordered={true} size="middle">
         <Descriptions.Item label="Date of Birth">{individual.date_of_birth || EM_DASH}</Descriptions.Item>
         <Descriptions.Item label="Sex">{individual.sex || "UNKNOWN_SEX"}</Descriptions.Item>
-        <Descriptions.Item label="Age">{individual.age || "UNKNOWN_AGE"}</Descriptions.Item>
+        <Descriptions.Item label="Age">{individual.age.age || "UNKNOWN_AGE"}</Descriptions.Item>
```
davidlougheed, line 14: we have nullish coalescing support in our js builds now, so you could do something like this to avoid a crash if a start/end age is passed in, with the `?.` and ?? operators:

```js
individual.age?.age ?? "UNKNOWN_AGE"
```

davidlougheed (approved):

PR #142, opened 2022-01-18
Fix for stale search results

gsfk:
When launching a new search, don't show old search results.  (Redmine Bug #953)


gsfk:
merged into #145

PR #143, opened 2022-01-18
Biosamples tab: add histological diagnosis

gsfk:
Add "histological_diagnosis" property to Biosamples overview


gsfk:
merged into #145

PR #148, opened 2022-03-07
improvements to "Other" pie chart category, plus bugfixes

gsfk:
Fixes for pie chart "Other" category:
- move "other" threshold control to header (Redmine #987)
- skip autoquery for the grouped "other" category (search does not work here, so no point offering search)
- don't use "Other" if there is only a single category below the threshold for grouping into "Other"

unrelated bugfixes:
- Remove double "variant" option from search 
- fix for crash where unindexed igv files were expected but missing
- fix for crash where experiments have no experiment results

### DIFF
--- a/src/components/OverviewContent.js
+++ b/src/components/OverviewContent.js
@@ -6,16 +6,10 @@ import ClinicalSummary from "./overview/ClinicalSummary";
 import VariantsSummary from "./overview/VariantsSummary";
 import { SITE_NAME } from "../constants";
 import ExperimentsSummary from "./overview/ExperimentsSummary";
-import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../constants";
 
 class OverviewContent extends Component {
     constructor(props) {
         super(props);
-        const threshold =
-      JSON.parse(localStorage.getItem("otherThresholdPercentage")) ?? DEFAULT_OTHER_THRESHOLD_PERCENTAGE;
-        this.state = {
-            otherThresholdPercentage: threshold,
-        };
     }
 
     componentDidMount() {
--- a/src/components/OverviewContent.js
+++ b/src/components/OverviewContent.js
@@ -37,9 +37,9 @@ class OverviewContent extends Component {
         </div>
         <Layout>
           <Layout.Content style={{ background: "white", padding: "32px 24px 4px" }}>
-            <ClinicalSummary otherThresholdPercentage={this.state.otherThresholdPercentage} />
+            <ClinicalSummary  />
             <Divider />
-            <ExperimentsSummary otherThresholdPercentage={this.state.otherThresholdPercentage} />
+            <ExperimentsSummary />
             <Divider />
             <VariantsSummary />
           </Layout.Content>
--- a/src/components/OverviewContent.js
+++ b/src/components/OverviewContent.js
@@ -6,7 +6,6 @@ import ClinicalSummary from "./overview/ClinicalSummary";
 import VariantsSummary from "./overview/VariantsSummary";
 import { SITE_NAME } from "../constants";
 import ExperimentsSummary from "./overview/ExperimentsSummary";
-import { Button } from "antd";
 import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../constants";
 
 class OverviewContent extends Component {
@@ -35,13 +34,6 @@ class OverviewContent extends Component {
           }}
         >
           <SitePageHeader title="Overview" style={{ border: "none" }} />
-          <Button
-            type="button"
-            icon="setting"
-            size={"small"}
-            style={{ alignSelf: "center", marginRight: "25px" }}
-            onClick={this.openModal}
-          />
         </div>
         <Layout>
           <Layout.Content style={{ background: "white", padding: "32px 24px 4px" }}>
--- a/src/components/OverviewContent.js
+++ b/src/components/OverviewContent.js
@@ -1,97 +1,57 @@
-import React, {Component} from "react";
-import {Layout, Divider} from "antd";
+import React, { Component } from "react";
+import { Layout, Divider } from "antd";
 
 import SitePageHeader from "./SitePageHeader";
 import ClinicalSummary from "./overview/ClinicalSummary";
 import VariantsSummary from "./overview/VariantsSummary";
-import OverviewSettingsControl from "./overview/OverviewSettingsControl";
 import { SITE_NAME } from "../constants";
 import ExperimentsSummary from "./overview/ExperimentsSummary";
-import { Modal, Button } from "antd";
+import { Button } from "antd";
 import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../constants";
 
 class OverviewContent extends Component {
     constructor(props) {
         super(props);
-        const startThreshold =
-          JSON.parse(localStorage.getItem("otherThresholdPercentage")) ?? DEFAULT_OTHER_THRESHOLD_PERCENTAGE;
+        const threshold =
+      JSON.parse(localStorage.getItem("otherThresholdPercentage")) ?? DEFAULT_OTHER_THRESHOLD_PERCENTAGE;
         this.state = {
-            otherThresholdPercentage: startThreshold,
-            previousThresholdPercentage: startThreshold,
-            modalVisible: false,
+            otherThresholdPercentage: threshold,
         };
-        this.setOtherThresholdPercentage = this.setOtherThresholdPercentage.bind(this);
-        this.openModal = this.openModal.bind(this);
-        this.setValueAndCloseModal = this.setValueAndCloseModal.bind(this);
-        this.cancelModal = this.cancelModal.bind(this);
     }
 
     componentDidMount() {
         document.title = `${SITE_NAME} - Overview`;
     }
 
-    setOtherThresholdPercentage(value) {
-        this.setState({otherThresholdPercentage: value});
-    }
-
-    openModal() {
-        this.setState({modalVisible: true});
-        // save threshold in case user cancels
-        this.setState({previousThresholdPercentage: this.state.otherThresholdPercentage});
-    }
-
-    setValueAndCloseModal() {
-        localStorage.setItem("otherThresholdPercentage", JSON.stringify(this.state.otherThresholdPercentage) );
-        this.setState({modalVisible: false});
-    }
-
-    cancelModal() {
-        this.setState({modalVisible: false});
-        this.setState({otherThresholdPercentage: this.state.previousThresholdPercentage});
-    }
-
     render() {
         return (
             <>
-            <div
-              style={{
-                  display: "flex",
-                  justifyContent: "space-between",
-                  background: "white",
-                  borderBottom: "1px solid rgb(232, 232, 232)",
-              }}
-            >
-              <SitePageHeader title="Overview" style={{ border: "none" }} />
-              <Button
-                type="button"
-                icon="setting"
-                size={"small"}
-                style={{ alignSelf: "center", marginRight: "25px" }}
-                onClick={this.openModal}
-              />
-            </div>
-            <Modal
-              visible={this.state.modalVisible}
-              closable={false}
-              destroyOnClose={true}
-              onOk={this.setValueAndCloseModal}
-              onCancel={this.cancelModal}
-            >
-              <OverviewSettingsControl
-                otherThresholdPercentage={this.state.otherThresholdPercentage}
-                setOtherThresholdPercentage={this.setOtherThresholdPercentage}
-                setValueAndCloseModal={this.setValueAndCloseModal}
-              />
-            </Modal>
-            <Layout>
-              <Layout.Content style={{ background: "white", padding: "32px 24px 4px" }}>
-                <ClinicalSummary otherThresholdPercentage={this.state.otherThresholdPercentage} />
-                <Divider />
-                <ExperimentsSummary otherThresholdPercentage={this.state.otherThresholdPercentage} />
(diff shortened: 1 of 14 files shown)


src/components/explorer/IndividualTracks.js
```diff
@@ -105,19 +105,19 @@ const IndividualTracks = ({individual}) => {
             return;
         }
 
-        if (!allTracks.length || !hasFreshUrls(allTracks, igvUrls) || igvRendered.current) {
+        if (!allFoundFiles.length || !hasFreshUrls(allTracks, igvUrls) || igvRendered.current) {
```
ppillot, line 108: I don't really know about this, but is the value for `allFoundFiles` updated dynamically? It seems to me that `allTracks` will change with the component whereas `allFoundFiles` is only set at initialization(?)
Could this have an impact on this test?

src/components/discovery/DiscoveryQueryBuilder.js
```diff
@@ -130,6 +130,7 @@ class DiscoveryQueryBuilder extends Component {
             <Menu onClick={this.handleAddDataTypeQueryForm}>
                 {this.props.servicesInfo
                     .filter(s => (this.props.dataTypes[s.id]?.items ?? []).length)
+                    .filter(s => s.name !== "Bento Variant Service")
```
ppillot, line 133: Nitpicking. This loops over the array another time. Can be combined within the previous filter using `&&`

src/components/explorer/IndividualExperiments.js
```diff
@@ -89,139 +89,120 @@ const IndividualExperiments = ({ individual }) => {
...
-      rowKey="filename"
-      dataSource={e.experiment_results.sort((r1, r2) => (r1.file_format > r2.file_format ) ? 1 : -1)}
-    />
-    {i !== (experimentsData.length - 1) && <Divider/>}
-  </div>
-))}
+          </div>
+          <div className="experiment-titles">
+            <Typography.Text style={titleStyle} level={4}>
+              {" "}
```
ppillot, line 189: Out if curiosity, could you clarify what this is for? (I believe it adds a regular space?)

src/components/SiteHeader.js
```diff
@@ -30,125 +36,158 @@ class SiteHeader extends Component {
...
 
         return (
-          <Layout.Header>
-            <Link to={BASE_PATH}>
-              <div
-                style={{
-                    margin: "0 15px 0 0",
-                    float: "left",
-                }}
+            <>
```
ppillot, line 140: I think this should be less indented as it wraps the markup that follows.

src/components/SiteHeader.js
```diff
@@ -30,125 +36,158 @@ class SiteHeader extends Component {
...
-            </Menu>
-          </Layout.Header>
+                {menuItems.map((i) => renderMenuItem(i))}
+              </Menu>
+            </Layout.Header>
+            <OverviewSettingsControl
+              modalVisible={this.state.modalVisible}
+              toggleModalVisibility={this.toggleModalVisibility}
+            />
+            </>
```
ppillot, line 170: One level less of indentation?

src/components/SiteHeader.js
```diff
@@ -30,125 +36,158 @@ class SiteHeader extends Component {
...
+                ),
                 onClick: () => this.props.showNotificationDrawer(),
-            }
+                key: "notifications",
+            },
+            {
+                style: { float: "right" },
+                icon: <Icon type="setting" />,
+                text: <span className="nav-text">Settings</span>,
+                onClick: this.toggleModalVisibility,                //xxxxxxxxx
```
ppillot: spurious comment to remove?

src/components/OverviewContent.js
```diff
@@ -1,97 +1,49 @@
...
-    }
-
-    cancelModal() {
-        this.setState({modalVisible: false});
-        this.setState({otherThresholdPercentage: this.state.previousThresholdPercentage});
-    }
-
     render() {
         return (
             <>
```
ppillot, line 21: Indentation: this markup wraps the following. It should be at least as indented (and I would say less indented)

src/components/OverviewContent.js
```diff
@@ -1,97 +1,49 @@
...
 import ExperimentsSummary from "./overview/ExperimentsSummary";
-import { Modal, Button } from "antd";
 import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../constants";
 
 class OverviewContent extends Component {
     constructor(props) {
         super(props);
-        const startThreshold =
-          JSON.parse(localStorage.getItem("otherThresholdPercentage")) ?? DEFAULT_OTHER_THRESHOLD_PERCENTAGE;
+        const threshold =
```
ppillot: Consider wrapping this in a try/catch as JSON.parse is susceptible to throw errors when the string is malformed. (in worst case scenario, this could prevent the application to start properly, even when the page is reloaded)

ppillot (changes requested):
I've went through the code. Mostly minor comments, please consider these in the light of my meagre knowledge of the codebase.
There might be some logic worth investigating in utils/mapNameValueFields.
I'll test the code later.

Many changes are whitelines changes. Is this due to the usage of a code formatter or are they manual additions?

src/utils/mapNameValueFields.js
```diff
@@ -1,21 +1,26 @@
-export const mapNameValueFields = (data, otherThreshold = 0.04) => {
+import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../constants";
+
+export const mapNameValueFields = (data, otherThreshold = DEFAULT_OTHER_THRESHOLD_PERCENTAGE) => {
```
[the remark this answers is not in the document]
gsfk: Thanks for this one! This is just a mistake on my part, although it shows the value of good variable naming.

src/utils/mapNameValueFields.js
```diff
@@ -1,21 +1,26 @@
...
 
     // Group the items in the array of objects denoted by
     // a "name" and "value" parameter
     const results = [];
     Object.entries(data).forEach(([key, val]) => {
-        // Group all elements with a small enough value together under an "Other"
-        if (val > 0 && (val / sumOfAllValues) < otherThreshold) {
+        // Group all elements with a small enough value together under an "Other", as long as there's more than one
+        if (numCategoriesBelowThreshold > 1 && val > 0 && (val / sumOfAllValues) < otherThreshold) {
             const otherIndex = results.findIndex(ob => ob.name === "Other");
```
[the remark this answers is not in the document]
gsfk: numCategoriesBelowThreshold is a bool, though, not a function call. Or is this a comment on the rest of the code?

src/modules/explorer/reducers.js
```diff
@@ -28,8 +31,10 @@ export const explorer = (
         isFetchingDownload: false,
 
         autoQuery: {
-            isAutoQuery: false
-        }
+            isAutoQuery: false,
+        },
+        otherThresholdPercentage:
+      JSON.parse(localStorage.getItem("otherThresholdPercentage")) ?? DEFAULT_OTHER_THRESHOLD_PERCENTAGE,
```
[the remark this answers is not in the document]
gsfk: Thanks, I'll write a proper function that does localStorage calls safely.

src/utils/mapNameValueFields.js
```diff
@@ -1,21 +1,26 @@
...
+import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../constants";
+
+export const mapNameValueFields = (data, otherThreshold = DEFAULT_OTHER_THRESHOLD_PERCENTAGE) => {
     if (!data)
         return [];
 
     // Accumulate all values to compute on them later
     const sumOfAllValues = Object.values(data).reduce((acc, v) => acc + v, 0);
+    const numCategoriesBelowThreshold = Object.values(data).filter(
+        (val) => val / sumOfAllValues < otherThreshold
```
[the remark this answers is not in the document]
gsfk: I'll attach most of the comments on `mapNameValueFields() `here: 

This is legacy code that has been built on top of a few times (including now!) that could probably benefit from a rewrite. The new code indeed loops twice through every array, but these are always small arrays; the alternative is a bunch of extra conditional handling inside the main loop. Again, a total rewrite could make this cleaner. 

Other issues are also a function of incremental changes elsewhere in the code. For example, `mapNameValueFields` is used only for pie charts, which don't even show categories with zero members, they are filtered out. So that could probably be done here instead. Similarly, a threshold parameter is always passed, so there's no real reason for the default assignment.

src/components/overview/OverviewSettingsControl.js
```diff
@@ -18,39 +29,68 @@ const OverviewSettingsControl = ({ otherThresholdPercentage, setOtherThresholdPe
...
+        toggleModalVisibility();
+    };
+
+    const cancelModal = () => {
+        setInputValue(previousThreshold);
+        setOtherThresholdPercentageInStore(previousThreshold);
+        toggleModalVisibility();
+    };
+
+    const writeThresholdToLocalStorage = (value) => {
```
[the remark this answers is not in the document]
gsfk: an alternative I considered here was a redux action creator (not a reducer!) that also handles writes to the store, so that everything is done in the same place. That's probably sensible.

src/components/overview/OverviewSettingsControl.js
```diff
@@ -1,14 +1,25 @@
...
-import { Col, InputNumber, Row, Slider } from "antd";
+import { useDispatch, useSelector } from "react-redux";
+import { Col, InputNumber, Row, Slider, Modal } from "antd";
 import PropTypes from "prop-types";
+import { setOtherThresholdPercentage } from "../../modules/explorer/actions";
+import { DEFAULT_OTHER_THRESHOLD_PERCENTAGE } from "../../constants";
 
+const OverviewSettingsControl = ({ modalVisible, toggleModalVisibility }) => {
+    const otherThresholdPercentage = useSelector((state) => state.explorer.otherThresholdPercentage);
+    const [inputValue, setInputValue] = useState(
```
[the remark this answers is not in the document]
gsfk: This is retrieved, not duplicated. It will update whenever the store updates.

https://react-redux.js.org/api/hooks#useselector

src/components/discovery/DiscoveryQueryBuilder.js
brouillette: I would add this new line filtering out Bento Variant Service won't be necessary after bentoV2 v2.6 release

src/components/explorer/IndividualTracks.js
gsfk: This particular test is fine, since we don't want to update `allFoundFiles` (they were found or they weren't) and the only check here is whether we found any files at all. This change addresses a bug where the igv viewer was launched even though there were no files to display.

`allTracks` _is_ mutated, but only properties of individuals tracks are changed (like whether they are viewable). In the long term I think the ideal fix is to remove the drs file searching responsibilities from this component entirely and put it in the parent component ExplorerIndividualContent,js.This will improve the overlap in responsibilities as well as eliminating some code duplication elsewhere.

src/components/discovery/DiscoveryQueryBuilder.js
gsfk: This is another readability vs clever one-liner issue. I usually opt for readability (and the array here is less than 10 items). Once the old variant service is removed this line can be removed entirely.

src/components/explorer/IndividualExperiments.js
gsfk: ah, this is a linter artifact which was already present in the code. Removed!

src/components/SiteHeader.js
gsfk: This, and all the other indentation issues, are a product of the linter settings. I agree they don't look great, but they're beyond the scope of this pr (and the pr won't even merge if the linter disagrees with you on the indentation)

src/components/SiteHeader.js
gsfk: gone!

gsfk:
apart from small fixes, the main changes I propose here are:
- rewrite `mapNameValueFields()` to better match how it's used
- safer handling of localStorage calls (I'll probably write a separate function for this)

src/utils/mapNameValueFields.js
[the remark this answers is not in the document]
ppillot: My point was about `results.findIndex()` which loops over the array to find an index which could be known by using a var instantiated outside the loop.
Something like 
```javascript
let otherIndex = -1;
const hasOtherCategory = numCategoriesBelowThreshold > 1;

Object.values(data).foreach((v) => {
    ...
    if (hasOtherCategory 
        && val > 0 
        && (val / sumOfAllValues) < otherThreshold
    ) {
            if (otherIndex === -1) {
                otherIndex = results.length;
                results.push({name: "Other", value: val, skipAutoquery: true}); // Create a new  element in the array
            } else {
                results[otherIndex].value += val; // Accumulate
            }
})
```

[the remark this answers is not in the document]
gsfk: ah, okay. This isn't my code. All the more reason for a rewrite.

src/components/OverviewContent.js
gsfk: addressed above

src/components/overview/OverviewSettingsControl.js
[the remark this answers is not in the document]
gsfk: I decided against always writing to redux and localStorage at the same time, simply because there's a performance hit with no benefit. At most we only need to write back to localStorage on close ("OK" or "Cancel").

ppillot (approved):
Nice!

PR #155, opened 2022-06-13
Experiments UI improvements

SanjeevLakhwani:
## Frontend Changes
- Experiments are now displayed in collapse
- When redirected to a specific experiment the experiment automatically opens up uncollapse
## Backend Changes
- added .idea to .gitignore
- added prettier config for jetbrain based environments
- converted clinical summary to a more dynamic, organised and smaller code
- changed how the redirection works from biosample to use experiment id as the key instead of biosample id (for cases of multiple experiments on a single biosample)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .idea/workspace.xml
 dist
 node_modules/
+.idea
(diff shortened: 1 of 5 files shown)


ppillot (approved):
Nice simplification of the code. Minor changes have been made after discussion. Testing has been successful.  Looks good for upcoming release, Thanks!

PR #160, opened 2022-07-20
Fixed csv export failing for individual search results

SanjeevLakhwani:
- The export failed as it made get request with all the requested ids in it, which exceeded the limit
- The changes were made to katsu to provide a post request endpoint for such a request
- This commit utilizes that change to make a post request to that end point and download csv

### DIFF
--- a/src/modules/explorer/actions.js
+++ b/src/modules/explorer/actions.js
@@ -52,12 +52,8 @@ export const performSearchIfPossible = (datasetID) => (dispatch, getState) => {
 
 export const performIndividualsDownloadCSVIfPossible = (datasetId, individualIds, allSearchResults) =>
     (dispatch, getState) => {
-        let ids;
-        if (individualIds.length > 0) { // Get only selected results
-            ids = individualIds;
-        } else { // Get all search results
-            ids = allSearchResults.map(sr => sr.key);
-        }
+
+        const ids = individualIds.length ? individualIds : allSearchResults.map(sr => sr.key);
 
         const myHeaders = new Headers();
         myHeaders.append("Content-Type", "application/json");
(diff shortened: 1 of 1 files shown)


src/modules/explorer/actions.js
```diff
@@ -57,21 +52,33 @@ export const performSearchIfPossible = (datasetID) => (dispatch, getState) => {
...
+        let ids;
         if (individualIds.length > 0) { // Get only selected results
-            dataUrl += ("&page_size=" + individualIds.length);
-            individualIds.forEach(id => dataUrl += `&id=${id}`);
+            ids = individualIds;
         } else { // Get all search results
-            dataUrl += ("&page_size=" + allSearchResults.length);
-            allSearchResults.forEach(sr => dataUrl += `&id=${sr.key}`);
+            ids = allSearchResults.map(sr => sr.key);
         }
```
ppillot, line 60: Nit. and personal preference: we could use a ternary operator here to declare and assign value to `ids` (as a const)

ppillot (commented):
Looks fine. Thanks!
We'll just make sure about the naming of the katsu endpoint before merging this.

src/modules/explorer/actions.js
```diff
@@ -2,6 +2,8 @@ import {createNetworkActionTypes, networkAction} from "../../utils/actions";
 import {jsonRequest} from "../../utils/requests";
 import {extractQueriesFromDataTypeForms} from "../../utils/search";
 
+import FileSaver from "file-saver";
+
 export const PERFORM_SEARCH = createNetworkActionTypes("EXPLORER.PERFORM_SEARCH");
 export const PERFORM_INDIVIDUAL_CSV_DOWNLOAD = createNetworkActionTypes("EXPLORER.PERFORM_INDIVIDUAL_CSV_DOWNLOAD");
```
gsfk, line 8: this is now dead code, and it's exported to reducers.js.

gsfk (commented):
There's several Bento utils for network calls that aren't used here (the "network actions" stuff in src/utils/actions.js), so they shouldn't be skipped without good reason. 

Also, I'm puzzled why there's an action creator but no action. Admittedly the old code made an action that was an impure function, so that's obviously not preferable.

src/modules/explorer/actions.js
SanjeevLakhwani: U

src/modules/explorer/actions.js
SanjeevLakhwani: My bad I forgot to ask about it. This part isn't completed fully yet and requires more discuss about post requests in this case.

PR #167, opened 2022-08-22
Enabled bigwigs download

SanjeevLakhwani:
Added `.bw` and `.bigwig` as an allowed file type


gsfk (commented):
Thanks, bigwigs now download correctly. 

Note though that this branch has errors upstream (autoquery is broken for experiments pie charts in overview)

PR #168, opened 2022-08-23
enabled inline source maps

SanjeevLakhwani:
- Enabled inline source maps in development mode


ppillot:
The tests are not conclusive at the moment
See in the screenshot bellow, the original source file is not displayed.
Screen Shot 2022-09-19 at 3 50 13 PM

PR #173, opened 2022-09-11
Dats correction

SanjeevLakhwani:
Added option to upload dats file

### DIFF
--- a/src/components/datasets/DatasetForm.js
+++ b/src/components/datasets/DatasetForm.js
@@ -16,7 +16,7 @@ const validateJson = (rule, value, callback) => {
     } catch (e) {
         callback("Please enter valid JSON");
     }
-}
+};
 
 const DatasetForm = ({ style, initialValue, form }) => {
     return (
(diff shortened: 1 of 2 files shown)


src/components/datasets/DatasetForm.js
```diff
@@ -9,26 +9,41 @@ import DataUseInput from "../DataUseInput";
...
                 })(<Input.TextArea placeholder="This is a dataset" />)}
             </Item>
             <Item label="Contact Information">
                 {form.getFieldDecorator("contact_info", {
-                    initialValue: (initialValue || { contact_info: "" }).contact_info || "",
+                    initialValue: initialValue?.contact_info || "",
                 })(<Input.TextArea placeholder={"Name\nInfo@c3g.ca"} />)}
             </Item>
+            <Item label="DATS File">
+                {form.getFieldDecorator("dats_File", {
```
ppillot, line 42: are you sure about the casing? `dats_File`? I would have expected `dats_file` as in the propTypes

ppillot (commented):
The code looks good (minor comment about casing on 1 variable)
~~The ticket mentioned, if time permit to allow uploading a file when a user has one available. Would it be feasible to add the control for it + the code to load the file content on the client side (it's a text format, so no need to do the whole file upload to backend ceremony)?~~ (I've just seen it's in another branch, not a PR though)

PR #174, opened 2022-09-22
Features/web registry version

noctillion:
-In  ServiceList.js, a component was added to render the information from environment mode, chord-services, and git. The additional information is displayed in the table in the version column beside the version information when the service is in dev mode when the command make run-web-dev is executed.


noctillion:
Discussed to PP, all agreed to merge!

PR #176, opened 2022-10-13
bugfix for correct search spinner behaviour

gsfk:
Fixes a minor annoyance with Bento search (Redmine [#1086](https://206.12.92.46/issues/1086)).

Currently when using either "Text Search" or "Advanced Search" both search buttons will spin while the search is waiting for results. This PR enforces the expected behaviour:

```
			Text Search Button	Advanced Search Button
during Text Search 	spinning		disabled
during Advanced Search	disabled		spinning
```

Other behaviours should be unchanged.

### DIFF
--- a/src/components/explorer/ExplorerDatasetSearch.js
+++ b/src/components/explorer/ExplorerDatasetSearch.js
@@ -107,9 +107,11 @@ class ExplorerDatasetSearch extends Component {
 
         const numResults = (this.props.searchResults || {searchFormattedResults: []}).searchFormattedResults.length;
 
+        const isFetchingSearchResults = this.props.fetchingSearch || this.props.fetchingTextSearch
+
         const tableStyle = {
-            opacity: (this.props.fetchingSearch || this.props.fetchingTextSearch ? 0.5 : 1),
-            pointerEvents: (this.props.fetchingSearch || this.props.fetchingTextSearch ? "none" : "auto")
+            opacity: (isFetchingSearchResults ? 0.5 : 1),
+            pointerEvents: (isFetchingSearchResults ? "none" : "auto")
         };
 
         // Calculate page numbers and range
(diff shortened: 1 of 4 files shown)


src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -108,8 +108,8 @@ class ExplorerDatasetSearch extends Component {
         const numResults = (this.props.searchResults || {searchFormattedResults: []}).searchFormattedResults.length;
 
         const tableStyle = {
-            opacity: (this.props.fetchingSearch ? 0.5 : 1),
-            pointerEvents: (this.props.fetchingSearch ? "none" : "auto")
+            opacity: (this.props.fetchingSearch || this.props.fetchingTextSearch ? 0.5 : 1),
+            pointerEvents: (this.props.fetchingSearch || this.props.fetchingTextSearch ? "none" : "auto")
```
davidlougheed, line 112: `this.props.fetchingSearch || this.props.fetchingTextSearch` is reused a couple times; maybe it could be assigned a name above this definition to reuse later in `render()`

src/components/explorer/SearchAllRecords.js
```diff
@@ -38,10 +38,13 @@ SearchAllRecords.propTypes = {
...
     searchAllRecords: PropTypes.object,
+    isFetchingAdvancedSearch: PropTypes.bool,
+    isFetchingTextSearch: PropTypes.bool
 };
 
-const mapStateToProps = (state) => ({
-    searchAllRecords: state.explorer,
+const mapStateToProps = (state, ownProps) => ({
+    isFetchingAdvancedSearch: state.explorer.fetchingSearchByDatasetID[ownProps.datasetID] || false,
+    isFetchingTextSearch: state.explorer.fetchingTextSearch || false,
```
davidlougheed, line 47: what is the `|| false` for here? 
also for defaulting `undefined` to `false` for the line above, I think `??` is a better operator to use

davidlougheed (changes requested):
a few changes requested but overall looks good!

src/components/explorer/SearchAllRecords.js
gsfk: This style is used throughout (see, eg, ExplorerDatasetSearch) so I'm probably guilty of following it without much thought. Switched to `??`

davidlougheed: yeah I think probably it's a pattern from pre-mainlining of `??` which happened in ~2020

davidlougheed (approved):
lgtm

PR #177, opened 2022-11-01
Remove peers/federation code

davidlougheed:
No longer used and even if we bring back federation it will probably be with a different approach.


gsfk (approved):
Looks good.

Is it worth the refactor to get rid of `modules/discovery`? There's only a single gohan feature in there, which probably belongs in `modules/explorer` anyhow.

Also... and not super-crucial... does Explorer still belong at `/data/explorer`? There's now nothing else at `/data`.

davidlougheed:
I agree that would be good. I'll merge this for now and open a Redmine ticket to do that refactor...

PR #189, opened 2022-12-06
Improve footer

davidlougheed:
* make component functional
* add `bento_web` version (imported from `package.json`) to footer
* adds spacing between logo and copyright text


gsfk (approved):
Looks good. There's a very minor bump to the igv version which I'm not able to test currently.

We've been inconsistent in the past with committing `package-lock.json`; this one isn't giving me any trouble. We've also previously overlooked changing the version number in `package.json` (this one is +0.0.1 ahead of the current version, I assume that's deliberate)

PR #191, opened 2022-12-08
Features/search sort order

gsfk:
- Bugfix for disappearing sort order for search results table ([Redmine #921](https://206.12.92.46/issues/921))

Previously when navigating away from the results page, any custom table sorting would be lost (table would default to sorting by ascending patient ids). This pr saves sorting preferences in redux.

### DIFF
--- a/src/components/explorer/ExplorerDatasetSearch.js
+++ b/src/components/explorer/ExplorerDatasetSearch.js
@@ -76,7 +76,7 @@ class ExplorerDatasetSearch extends Component {
 
         const {sortColumnKey, sortOrder} = this.props.tableSortOrder;
 
-        const SEARCH_RESULT_COLUMNS = [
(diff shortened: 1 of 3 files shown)


src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -99,6 +73,39 @@ class ExplorerDatasetSearch extends Component {
     }
 
     render() {
+
+        const {sortColumnKey, sortOrder} = this.props.tableSortOrder;
+
+        const SEARCH_RESULT_COLUMNS = [
```
davidlougheed, line 79: change this to camel case since it's no longer global const for the file. other than that looks good!

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #195, opened 2022-12-14
Runtime config

davidlougheed:
Will support dockerization efforts...


brouillette (approved):
looks good!
works locally and on staging

PR #197, opened 2022-12-14
Features/igv save position

gsfk:
Store last IGV position in redux. 

When tabbing between "Tracks" and other parts of individual overview (Biosamples, Experiments, etc), the position of the IGV browser will be preserved, instead of defaulting to the "zoomed out" view. Position will also persist across individuals, unless a hard refresh occurs. 

A few subtleties are worth noting:

- position is monitored by listening to IGV's `locuschange` event. This can produce a very large number of calls, so storing every change degrades performance. Instead, lodash `debounce` is used to handle only the last event over a particular time frame (currently set to 500 ms).

- igv position is retrieved from the redux store only on the first render.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,7 @@
     "file-saver": "^2.0.5",
     "igv": "^2.13.6",
     "iso8601-duration": "^1.2.0",
+    "lodash.debounce": "^4.0.8",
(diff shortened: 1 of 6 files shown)


davidlougheed (approved):
looks good to me

PR #199, opened 2022-12-23
feature 1400

brouillette:
Explorer -> Variant search: Assembly combobox now only displays choices for which data exist inside of Gohan


src/components/discovery/VariantSearchHeader.js
```diff
@@ -2,8 +2,12 @@ import React, { useState } from "react";
 import { Form, Select } from "antd";
 import PropTypes from "prop-types";
 import LocusSearch from "./LocusSearch";
+import { useSelector } from "react-redux";
 
 const VariantSearchHeader = ({dataType, addVariantSearchValues}) => {
+    const varOvRes = useSelector((state) => state.explorer.variantsOverviewResponse);
+    const ovAsmIds = varOvRes?.assemblyIDs !== undefined ? Object.keys(varOvRes?.assemblyIDs) : [];
```
gsfk, line 9: variable names here are obscure to me

gsfk (approved):
Looks good! A nice-to-have feature here would be that, in cases where there is only one assembly, it is already selected. This saves the user a pointless click, and gives more information than an empty box. Could always be another ticket I suppose. 

I also quibbled with some variable names, but feel free to take that with a grain of salt.

src/components/discovery/VariantSearchHeader.js
brouillette: I agree
The linter doesn't like lines longer than 120 characters, so I abbreviated them

src/components/discovery/VariantSearchHeader.js
```diff
@@ -17,18 +21,6 @@ const VariantSearchHeader = ({dataType, addVariantSearchValues}) => {
...
-    // files ingested with GRCh37 reference need to be searched using assemblyId "Other"
-
-    // so if assembly is "Other", pass that value to the form, but use
-    // "GRCh37" as the reference for gene lookup in Gohan
-        if (value === "Other") {
-            setLookupAssemblyId("GRCh37");
-            return;
-        }
-
-        setLookupAssemblyId(value);
```
gsfk, line 31: setLookupAssemblyId(value) needs to stay or variant search will break. 

only  the` if (value === "Other")` block needs to go.

brouillette: fixed

src/components/discovery/VariantSearchHeader.js
brouillette: see https://github.com/bento-platform/bento_web/pull/199/commits/d70960701bfb1b8a67ef59bbfe7d0c3f336471cc for the nice-to-have!

PR #202, opened 2023-01-04
Fix backwards compatibility for old Dockerfiles / bentoV2 versions

davidlougheed:
chore: fix back-compat for old dockerfiles/bentoV2
chore: update minor dependencies


gsfk:
Yes, nice, this seems to work well, thanks. I assume the Dockerfile changes are to keep forwards rather than backwards compatibility, since this Dockerfile is ignored in 2.8.

PR #204, opened 2023-01-05
[WIP] feature 1401

brouillette:
Explorer -> Variant search: Remove search conditions "start", "end", "sample_id"

### DIFF
--- a/src/components/discovery/DiscoverySearchForm.js
+++ b/src/components/discovery/DiscoverySearchForm.js
@@ -390,17 +390,17 @@ class DiscoverySearchForm extends Component {
             >
                 {this.state.isVariantSearch ?
                     <></> :
-                this.state.isPhenopacketSearch ? (
+                    this.state.isPhenopacketSearch ? (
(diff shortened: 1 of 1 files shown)


davidlougheed:
we can go this direction as a temp hack for 2.10, but in the long run i'd like to separate the "add condition"-type search completely from the "user friendly" search inputs, since this gives us complete flexibility with the UI. I'm a bit hesitant to always stack hacks onto UI code since it makes it hard to reason with, but if this is a must-have we can do it this way... to be discussed on monday i guess!

gsfk (approved):
Honestly I think it's sensible as a temporary measure: there aren't really any conditions to add, so no real call for having an "add condition" button. It could be more clear which fields are optional and which aren't, although that's probably a separate ticket.

For the future we probably need to move past the approach of having the search interface just be a mapping from the schema... it's super-flexible, but not super-intuitive to use.

davidlougheed (approved):

PR #208, opened 2023-01-20
Features/variant search improvements

gsfk:
This PR addresses multiple issues in the variants search UI, including accommodations for new issues (Redmine 
[1400](https://206.12.92.46/issues/1400), [1401](https://206.12.92.46/issues/1401), [1402](https://206.12.92.46/issues/1402)) along with some older headaches ([1404](https://206.12.92.46/issues/1404), [957](https://206.12.92.46/issues/957))

Large changes: 
- Variants search UI now handles optional fields. Any optional field left unfilled when search is launched will be ignored.
- `Genotype` field is now optional, so can be left empty. It is now also possible to make a selection but then remove it (with the "x" at the right edge of the input box).
- `Reference` and `Alternate` fields can be filled or left blank. Note that they only accept `ACGT/acgt`and `backspace` and will give error feedback for anything else. (This was added in pr #199).
- There is now some form verification for variants (Redmine [#1404](https://206.12.92.46/issues/1404)). Previously when a form was incorrectly filled no feedback was given.  User feedback is given on submit, there are still edge cases (such as multiple repeated and edited searches) where it appears earlier. 
- There is now also feedback when you tab away from the `Gene / position` field without making a correct entry. This is a partial solution to Redmine [#957](https://206.12.92.46/issues/957). A full solution to that issue is probably possible only by replacing antd's Autocomplete with one from a different library. 

Also:
- Fixed a bug in the legacy code that would generate duplicate form keys in some circumstances, leading to weird behaviour when deleting search conditions and adding new ones. 

### Notes

- Requires latest Gohan to test
- Owing to a [Prettier bug](https://github.com/prettier/prettier-vscode/issues/2041), there are several useless whitespace changes, I have not had time to revert them all. 
- ... and finally, this is not perfect code. Most of the problems grow out of the design issues: having a user-facing form that is not actually submitted, but instead controls fields in a different form that is hidden from the user. This is not ideal, and not an edge case that UI libraries will bother with. In the long term a better solution is to have a distinct "user-friendly" search interface that is independent of the existing one, rather than bolted onto it. This will be easier to use and easier to code.

### DIFF
--- a/src/components/discovery/DiscoveryQueryBuilder.js
+++ b/src/components/discovery/DiscoveryQueryBuilder.js
@@ -86,7 +86,7 @@ class DiscoveryQueryBuilder extends Component {
 
     handleSubmit = async () => {
         // signal to variants components, may work better in redux
-        this.setState({isSubmitting: true})                 
+        this.setState({isSubmitting: true});
 
         try {
             await Promise.all(Object.entries(this.forms).filter(f => f[1]).map(([_dt, f]) =>
@@ -103,7 +103,7 @@ class DiscoveryQueryBuilder extends Component {
             (this.props.onSubmit ?? nop)();
         } catch (err) {
             console.error(err);
-            this.setState({isSubmitting: false})
+            this.setState({isSubmitting: false});
         }
     };
 
--- a/src/components/discovery/DiscoveryQueryBuilder.js
+++ b/src/components/discovery/DiscoveryQueryBuilder.js
@@ -85,7 +85,9 @@ class DiscoveryQueryBuilder extends Component {
     }
 
     handleSubmit = async () => {
-        this.setState({isSubmitting: true});
+        // signal to variants components, may work better in redux
+        this.setState({isSubmitting: true})                 
+
         try {
             await Promise.all(Object.entries(this.forms).filter(f => f[1]).map(([_dt, f]) =>
                 new Promise((resolve, reject) => {
@@ -101,6 +103,7 @@ class DiscoveryQueryBuilder extends Component {
             (this.props.onSubmit ?? nop)();
         } catch (err) {
             console.error(err);
+            this.setState({isSubmitting: false})
         }
     };
 
--- a/src/components/discovery/DiscoveryQueryBuilder.js
+++ b/src/components/discovery/DiscoveryQueryBuilder.js
@@ -18,7 +18,8 @@ class DiscoveryQueryBuilder extends Component {
         super(props);
 
         this.state = {
-            schemasModalShown: false
+            schemasModalShown: false,
+            isSubmitting: false
         };
 
         this.handleSubmit = this.handleSubmit.bind(this);
@@ -84,6 +85,7 @@ class DiscoveryQueryBuilder extends Component {
     }
 
     handleSubmit = async () => {
+        this.setState({isSubmitting: true});
         try {
             await Promise.all(Object.entries(this.forms).filter(f => f[1]).map(([_dt, f]) =>
                 new Promise((resolve, reject) => {
@@ -155,6 +157,7 @@ class DiscoveryQueryBuilder extends Component {
                     wrappedComponentRef={form => this.forms[id] = form}
                     onChange={fields => this.handleFormChange(dataType, fields)}
                     handleVariantHiddenFieldChange={this.handleVariantHiddenFieldChange}
+                    isSubmitting={this.state.isSubmitting}
                 />
             </Tabs.TabPane>;
         });
--- a/src/components/discovery/DiscoverySearchForm.js
+++ b/src/components/discovery/DiscoverySearchForm.js
@@ -14,9 +14,21 @@ import {
 import DiscoverySearchCondition, {getSchemaTypeTransformer} from "./DiscoverySearchCondition";
 import VariantSearchHeader from "./VariantSearchHeader";
 
-const NUM_HIDDEN_VARIANT_FORM_ITEMS = 7;
 const TOOLTIP_DELAY_SECONDS = 0.8;
 
+// required by ui not precisely the same as required in spec, possible TODO
+const VARIANT_REQUIRED_FIELDS = [
+    "[dataset item].assembly_id",
+    "[dataset item].chromosome",
+    "[dataset item].start",
+    "[dataset item].end",
+];
+
+const VARIANT_OPTIONAL_FIELDS = [
+    "[dataset item].calls.[item].genotype_type",
+    "[dataset item].alternative",
+    "[dataset item].reference",
+];
 
 // noinspection JSUnusedGlobalSymbols
 const CONDITION_RULES = [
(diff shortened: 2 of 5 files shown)


src/components/discovery/DiscoverySearchForm.js
```diff
@@ -306,103 +315,136 @@ class DiscoverySearchForm extends Component {
...
+                    {subjectOptions.map((o) => (
+                        <Menu.Item key={o.path}>
+                            <DropdownOption option={o} />
+                        </Menu.Item>
+                    ))}
                 </Menu.SubMenu>
-                <Menu.SubMenu title={<span style={{marginRight: "10px"}}>Phenotypic Features </span>}>
-                {phenotypicFeaturesOptions.map(o => <Menu.Item key={o.path}><DropdownOption option={o}/></Menu.Item>)}
+                <Menu.SubMenu title={<span style={{ marginRight: "10px" }}>Phenotypic Features </span>}>
+                    {phenotypicFeaturesOptions.map((o) => (
```
davidlougheed, line 327: the mapped function is repeated ~6 times, probably worth factoring it out?

src/components/discovery/LocusSearch.js
```diff
@@ -59,9 +74,10 @@ const LocusSearch = ({assemblyId, addVariantSearchValues}) => {
             return;
         }
 
-    // don't use locus.assemblyId, since this is the lookup value
         const {chrom, start, end} = locus;
         addVariantSearchValues({chrom: chrom, start: start, end: end});
```
davidlougheed, line 78: this can be shortened to `{chrom, start, end}` - and the same object is used for `setCurrentLocus` as well

i mention this again below, but this "make reduced locus form from locus" pattern appears twice and could probably be factored out; that function could describe why only these fields are included in this version of the locus object.

src/components/discovery/LocusSearch.js
```diff
@@ -25,13 +27,16 @@ const LocusSearch = ({assemblyId, addVariantSearchValues}) => {
...
+            setCurrentLocus({chrom: null, start: null, end: null});
+            setLocusValidity(false);
             return;
         }
 
         const chrom = result[1].toUpperCase(); //for eg 'x', has no effect on numbers
         const start = Number(result[2]);
         const end = Number(result[3]);
 
+        setCurrentLocus({chrom: chrom, start: start, end: end});
```
davidlougheed, line 39: this can be shortened to `{chrom, start, end}` - and the same object is used for `addVariantSearchValues` as well

src/components/discovery/VariantSearchHeader.js
```diff
@@ -6,40 +6,92 @@ import { useSelector } from "react-redux";
...
+
+        if (isValidLocus(locus)) {
+            setLocusValidity(true);
+
+        } else {
+            setLocusValidity(false);
+            return;
+        }
+
+        setLocus({chrom: chrom, start: start, end: end});
```
davidlougheed, line 89: this pattern of constructing a reduced form of the locus (locus, start, end) seems to pop up a few times - could be worth making a function to build it.

also, this again could be written more compactly as `{chrom, start, end}`

davidlougheed (changes requested):
some stylistic suggestions; haven't actually ran this yet...

src/components/discovery/DiscoverySearchForm.js
gsfk: this is one of several useless whitespace changes noted above. This code is ancient!

src/components/discovery/LocusSearch.js
gsfk: honestly I always found this style confusing

src/components/discovery/DiscoverySearchForm.js
davidlougheed: oops

src/components/discovery/DiscoveryQueryBuilder.js
```diff
@@ -99,6 +103,7 @@ class DiscoveryQueryBuilder extends Component {
             (this.props.onSubmit ?? nop)();
         } catch (err) {
             console.error(err);
+            this.setState({isSubmitting: false});
```
v-rocheleau, line 106: I think that we should add a `finally` statement in order to set isSubmitting back to false when no errors are caught.

gsfk: the `isSubmitting` signal is really one of the weakest parts of the code.... if it's immediately changed back will anything at all happen? I suspect it will get better performance in redux rather than as a prop.

src/components/discovery/DiscoverySearchForm.js
```diff
@@ -30,10 +42,13 @@ const CONDITION_RULES = [
             const isEnum = value.fieldSchema.hasOwnProperty("enum");
 
             // noinspection JSCheckFunctionSignatures
-            if (searchValue === null
-                    || (!isEnum && !searchValue)
-                    || (isEnum && !value.fieldSchema.enum.includes(searchValue))) {
-                cb("This field is required.");
+            if (
+                !VARIANT_OPTIONAL_FIELDS.includes(value.field) &&
+                (searchValue === null ||
```
v-rocheleau, line 47: Mix of `searchValue === null` and `!searchValue` in condition chain, which can have different results (e.g. searchValue is undefined), is this intended? If not we could stick to `!searchValue`.

gsfk: everything below `!VARIANT_OPTIONAL_FIELDS.includes(value.field) &&` is legacy code, so I don't have an immediate answer. It's possible that some search values could be "0".

src/components/discovery/DiscoverySearchForm.js
```diff
@@ -293,9 +300,11 @@ class DiscoverySearchForm extends Component {
         // eslint-disable-next-line react/prop-types
         const DropdownOption = ({option: {path, ui_name: uiName}}) => {
```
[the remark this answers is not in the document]
gsfk: Happily (or not!) any of the components using antd `Form` are just far too convoluted for it to be worth converting them to functions.

src/components/discovery/DiscoveryQueryBuilder.js
v-rocheleau: Yeah having it stored in redux would help with props drilling performance issues I think.
As far as having isSubmitting immediately set to false after a successful submit (whether with component of redux state), nothing will happen right now as far as I can see, but there could be side effects in the future if we rely on this state value.

davidlougheed (approved):
looks good, will need merge conflict fixing

brouillette:
dealt with merge conflict by reverting the last 1401 pr merge that contained redundant code already handled in this branch

v-rocheleau (approved):
Looks good to me. But looking at the build.yml file I am confused as to why the build-push job was ran here with the `pull_request` trigger, since master is not the target branch.

brouillette:
> Looks good to me. But looking at the build.yml file I am confused as to why the build-push job was ran here with the `pull_request` trigger, since master is not the target branch.

Might be a side effect to my commit that I tried to revert https://github.com/bento-platform/bento_web/commit/16927c63edfdf1aee2f7abca3487c4e1a96ebfc1

PR #211, opened 2023-01-25
Patch variant autocomplete

gsfk:
Improved onBlur handling for Gene / position input in variants search: track current input value in component state and re-verify when tabbing away from input. 

Should resolve remaining issues with Redmine [#1404](https://206.12.92.46/issues/1404)

### DIFF
--- a/src/components/discovery/LocusSearch.js
+++ b/src/components/discovery/LocusSearch.js
@@ -26,7 +26,6 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
         const result = parse.exec(value);
 
         if (!result) {
-            setLocusValidity(false);
             return {chrom: null, start: null, end: null};
         }
 
--- a/src/components/discovery/LocusSearch.js
+++ b/src/components/discovery/LocusSearch.js
@@ -98,8 +98,7 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
             return;
         }
 
-        const {chrom, start, end} = locus;
-        addVariantSearchValues({chrom, start, end});
+        addVariantSearchValues(locus);
         handleLocusChange(locus);
     };
 
@@ -125,7 +124,7 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
             value={`${g.name}_${g.assemblyId}_autocomplete_option`}
 
             label={`${g.name} chrom: ${g.chrom}`}
-            locus={g}
+            locus={{"chrom": g.chrom, "start": g.start, "end": g.end}}
             // style={optionStyle}
           >{geneDropdownText(g)}</Option>
         ))}
--- a/src/components/discovery/LocusSearch.js
+++ b/src/components/discovery/LocusSearch.js
@@ -34,7 +34,7 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
         const start = Number(result[2]);
         const end = Number(result[3]);
 
-        addVariantSearchValues({chrom: chrom, start: start, end: end});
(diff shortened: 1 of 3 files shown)


src/components/discovery/LocusSearch.js
```diff
@@ -62,11 +66,32 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
...
+        const isPositionNotation = inputValue.includes(":");
+
+        if (!(isAutoCompleteOption || isPositionNotation)) {
+            handleLocusChange({chrom: null, start: null, end: null});
+            addVariantSearchValues({chrom: null, start: null, end: null});
+            return;
+        }
+
+        if (isPositionNotation) {
+            const {chrom, start, end} = parsePosition(inputValue);
```
davidlougheed, line 87: parsePosition always returns `{chrom, start, end}` shape, so there's not too much point in deconstructing & then reconstructing the object immediately for handleLocusChange/addVariantSearchValues IMO

src/components/discovery/LocusSearch.js
```diff
@@ -27,24 +26,29 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
...
+            return {chrom: null, start: null, end: null};
         }
 
         const chrom = result[1].toUpperCase(); //for eg 'x', has no effect on numbers
         const start = Number(result[2]);
         const end = Number(result[3]);
 
-        setCurrentLocus({chrom: chrom, start: start, end: end});
-        addVariantSearchValues({chrom: chrom, start: start, end: end});
+        return {chrom, start, end};
```
davidlougheed, line 37: I like this change; moving this function towards a more side-effect-free style

src/components/discovery/LocusSearch.js
```diff
@@ -27,24 +26,29 @@ const LocusSearch = ({assemblyId, addVariantSearchValues, handleLocusChange, set
         const result = parse.exec(value);
 
         if (!result) {
-            setCurrentLocus({chrom: null, start: null, end: null});
             setLocusValidity(false);
```
davidlougheed, line 29: given the returns not actually mutating the locus now, should this validity set still be here? elsewhere it seems like `setLocusValidity` is being called outside `parsePosition`, and from an code comprehension POV it seems that making parsePosition not have any state side effects would be an improvement - then, this function could also be moved outside of `LocusSearch` to avoid redefining it every render

davidlougheed (changes requested):
looks good, a few things I'm not sure about but other than that I think it improves the readability of these areas too!

src/components/discovery/LocusSearch.js
gsfk: yes, this side effect probably shouldn't be here. I think in fact it's redundant now that null values are passed to the parent component and then checked there.

src/components/discovery/LocusSearch.js
davidlougheed: yeah in this case parsePosition doesn't return an assemblyID though right?

gsfk: yes true

brouillette:
tested [sha-3c38151](https://github.com/orgs/bento-platform/packages/container/bento_web/66101040?tag=sha-3c38151) [pr-211](https://github.com/orgs/bento-platform/packages/container/bento_web/66101040?tag=pr-211) locally
works as expected!

davidlougheed (approved):

src/components/discovery/LocusSearch.js
gsfk: Fixed this one, and restructured locus to pull out another one.

src/components/discovery/LocusSearch.js
gsfk: gone

PR #216, opened 2023-02-28
cBioPortal integration + analysis runs + DRS object explorer

davidlougheed:


src/components/App.js
```diff
@@ -88,6 +89,8 @@ class App extends Component {
     }
 
     render() {
+        const margin = window.location.pathname.endsWith("cbioportal") ? 0 : 26;
```
v-rocheleau, line 92: I guess we change the margin to give some air to the cbioportal iframe?

src/components/manager/ManagerIngestionContent.js
```diff
@@ -13,233 +13,156 @@ import {
...
+    const {workflows, workflowsLoading} = useSelector(workflowsStateToPropsMixin);
+    const {selectedTable} = values;
+
+    const workflowItems = workflows.ingestion
+        .filter(w => w.data_type === (selectedTable ? selectedTable.split(":")[1] : null))
+        .map(w =>
+            <WorkflowListItem
+                key={w.id}
+                workflow={w}
+                selectable={true}
```
v-rocheleau, line 36: Just for my understanding, what makes some WorkflowListItems selectable or not?

davidlougheed: in the workflow tab, there are lists of workflows which cannot be clicked, whereas here clicking on one selects it as the ingestion workflow. probably, this should just be done based on the presence of a handler function rather than a separate attr, but this is just ported from old code right now.

davidlougheed: removed it in favour of just detecting the presence of a click handler

src/components/App.js
davidlougheed: added a clarifying comment

src/components/manager/ManagerIngestionContent.js
```diff
@@ -13,233 +13,155 @@ import {
...
+                selectedWorkflow={selectedWorkflow}
+                inputs={inputs}
+                handleRunWorkflow={handleRunWorkflow}
+            />
+        )}
+        onSubmit={({workflowSelectionValues, selectedWorkflow, inputs}) => {
+            const {selectedTable} = workflowSelectionValues;
+
+            if (!selectedTable || !selectedWorkflow) {
+                // TODO: GUI error message
```
v-rocheleau, line 148: Ok if not in this version? Later we could implement an error display system with redux, so we can easilly show GUI error messages by using dispatch.

davidlougheed: there's one built into ant design, i'll just add one now

davidlougheed: done

src/components/manager/RunSetupInputForm.js
v-rocheleau, line 1: Nice use of useCallback, was it added as a precaution or to fix a specific issue?

davidlougheed: no issue specifically, just trying to prevent needless re-renders since it's passed down

...which makes me realize it isn't needed for `getInputComponent`, just `handleSubmit`...

noctillion (approved):
I tested the changes in the code and couldn't find any evident issues. Great job!

PR #224, opened 2023-04-05
fix: remove CHORD mentions from UI and README

v-rocheleau:
Small patch for ticket 1519
- Replace CHORD wording in the interface to use "Bento"


davidlougheed (changes requested):
Bento isn't an acronym like CHORD was; use Bento instead of BENTO

davidlougheed (approved):
lgtm

PR #225, opened 2023-04-05
feat: drop box pdf viewer + file load error handling

davidlougheed:


gsfk (commented):
Looks good with just two issues: 

- breaks on file names with spaces (and presumably any other non-url character)
- when _does_ break, the spinner runs forever 

I don't see a problem with forbidding spaces in filenames, but the failure mode could be nicer

davidlougheed:
good catch, will figure something out!

davidlougheed:
this led me down a rabbit hole-- this is caused by a bug in gateway configs and a bug in drop box, but the web part is actually not at fault

davidlougheed:
https://github.com/bento-platform/bentoV2/pull/118 PR 1 related to this

davidlougheed:
https://github.com/bento-platform/bento_drop_box_service/pull/24 PR 2

PR #226, opened 2023-04-06
Fix/prop types err ingest

davidlougheed:
Fixes a prop type error on the 'run ingestion workflow' page


v-rocheleau (approved):
Lgtm. FYI I marked 1094 as ready-to-release too soon, thought the ticket was for lint errors.

PR #232, opened 2023-05-02
[12] Features/explorer biosample experiment table

noctillion:
Refactor ExplorerDatasetSearch.js to a Functional Component and Add Tabs for Individuals, Experiments, and Biosamples

This PR includes updates to the ExplorerDatasetSearch.js component. The primary goal of this PR is to refactor the ExplorerDatasetSearch.js from a class component to a functional component also introduces the following features:

-Tabs for Displaying Tables: The ExplorerDatasetSearch component now includes tabs for displaying separate tables for Individuals, Biosamples and Experiments. This allows to navigate the different data types and view relevant information for each category.

-Functional Components for Each Table: To support the new tabbed interface, it was added functional components for each table (IndividualsTable, ExperimentsTable, and BiosamplesTable). These components are responsible for structuring and displaying the corresponding data.

-Modifying the reducer and actions in Explorer to reshape the query results and make them available for both Biosamples and Experiment tables.


src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,165 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Link } from "react-router-dom";
+import { withBasePath } from "../../utils/url";
+import ExplorerSearchResultsTableComp from "./ExplorerSearchResultsTableComp";
+
+const biosampleRender = (individual, record) => {
+    const alternateIds = individual.alternate_ids ?? [];
+    const listRender = alternateIds.length ? " (" + alternateIds.join(", ") + ")" : "";
```
v-rocheleau, line 9: Tough to read with all the quotes, a [string template](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) would be a better option here. There is also a white space before the parenthesis in " ("

src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,165 @@
+import React from "react";
+import PropTypes from "prop-types";
+import { Link } from "react-router-dom";
+import { withBasePath } from "../../utils/url";
+import ExplorerSearchResultsTableComp from "./ExplorerSearchResultsTableComp";
+
+const biosampleRender = (individual, record) => {
```
v-rocheleau, line 7: Deconstruct and specify props

src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,165 @@
...
+    if (customPluralForms[word]) {
+        return customPluralForms[word];
+    } else if (word.slice(-1) !== "s") {
+        return word + "s";
+    }
+
+    return word;
+};
+
+const experimentsRender = (studiesType) => {
```
v-rocheleau, line 41: Define props types

src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,165 @@
...
+                </>
+            ) : (
+                <>—</>
+            )}
+        </>
+    );
+};
+
+const experimentsSorter = (a, b) => {
+    const countExperiments = (studiesType) => {
```
v-rocheleau, line 64: This function could be applied to any list, not just experiments, so we could put it in the utils file as `nonNullLenght`to be more generic

src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,165 @@
...
+        return studiesType.filter((s) => s !== null).length;
+    };
+
+    return countExperiments(a.study_types) - countExperiments(b.study_types);
+};
+
+const sampledTissuesRender = (sampledTissues) => {
+    return sampledTissues.map((m) => m.label)[0];
```
v-rocheleau, line 72: We pass a list as an argument, map all its items, but the function only returns index 0.

src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,165 @@
...
+};
+
+const sampledTissuesSorter = (a, b) => {
```
v-rocheleau, line 75: Only compare index 0?

v-rocheleau (changes requested):
I know it is still in draft but I took a quick look and made some comments

src/modules/explorer/reducers.js
```diff
@@ -245,6 +288,86 @@ export const explorer = (
...
+                }
+                return objects;
+            }, []);
+        })
+        .filter((entry) => entry.key !== null && entry.key !== undefined);
+}
+
+const tableSearchResultsBiosamples = (searchResults) => {
+    const res = generateBiosampleObjects(searchResults);
+    return res;
```
davidlougheed, line 369: this function does nothing except call `generateBiosampleObjects` which means you don't need `tableSearchResultsBiosamples` at all

src/modules/explorer/reducers.js
```diff
@@ -83,6 +85,8 @@ export const explorer = (
                     [action.datasetID]: {
                         results: action.data,
                         searchFormattedResults: tableSearchResults(action.data),
+                        searchFormattedResultsExperiment: tableSearchResultsExperiments(action.data),
+                        searchFormattedResultsBioSamples: tableSearchResultsBiosamples(action.data),
```
davidlougheed, line 89: we generally case this as `Biosamples` not `BioSamples`

src/modules/explorer/actions.js
```diff
@@ -55,29 +59,72 @@ export const performSearchIfPossible = (datasetID) => (dispatch, getState) => {
...
+    (datasetId, individualIds, allSearchResults) => (dispatch, _getState) => {
+        const ids = individualIds.length ? individualIds : allSearchResults.map((sr) => sr.key);
         return dispatch(performIndividualsDownloadCSV(ids));
     };
 
+// Action to perform the request to download biosamples CSV
+const performBiosamplesDownloadCSV = networkAction((ids) => (dispatch, getState) => ({
+    types: PERFORM_BIOSAMPLE_CSV_DOWNLOAD,
+    url: `${getState().services.itemsByArtifact.metadata.url}/api/batch/biosamples`,
+    req: jsonRequest(
```
davidlougheed, line 89: this `jsonRequest` function is called the same way in several places, so refact either this call into a function or the whole CSV request builder

davidlougheed (changes requested):

src/components/explorer/BiosamplesTable.js
noctillion: I think  accessing the label of the first item in the array is reasonable here, because all the objects in the array have the same label

src/components/explorer/BiosamplesTable.js
noctillion: the parenthesis is ok, changes to string template done!

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -1,265 +1,123 @@
...
-        console.log("search results: ", this.props.searchResults);
+    const hasIndividuals = searchResults && searchResults.searchFormattedResults;
+    const hasExperiments =
+        searchResults &&
+        searchResults.searchFormattedResultsExperiment &&
+        searchResults.searchFormattedResultsExperiment.length > 0;
+    const hasBiosamples =
+        searchResults &&
+        searchResults.searchFormattedResultsBiosamples &&
+        searchResults.searchFormattedResultsBiosamples.length > 0;
```
v-rocheleau, line 82: Here and above (line 78) we can simplify the evaluation by removing `> 0`, since 0 is [falsy](https://developer.mozilla.org/en-US/docs/Glossary/Falsy)

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -1,265 +1,123 @@
...
+                searchLoading={fetchingSearch}
+                addDataTypeQueryForm={(form) => dispatch(addDataTypeQueryForm(dataset, form))}
+                updateDataTypeQueryForm={(index, form) => dispatch(updateDataTypeQueryForm(dataset, index, form))}
+                removeDataTypeQueryForm={(index) => dispatch(removeDataTypeQueryForm(dataset, index))}
+            />
+            {hasIndividuals &&
+                !isFetchingSearchResults &&
+                (showTabs ? (
+                    <Tabs defaultActiveKey="1" onChange={onTabChange} activeKey={activeKey}>
+                        <TabPane tab="Individual" key="1">
```
v-rocheleau, line 102: Reference `TAB_KEYS.INDIVIDUAL` property for `key` prop.

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -1,265 +1,123 @@
...
+                onSubmit={resetPageNumber}
+                searchLoading={fetchingSearch}
+                addDataTypeQueryForm={(form) => dispatch(addDataTypeQueryForm(dataset, form))}
+                updateDataTypeQueryForm={(index, form) => dispatch(updateDataTypeQueryForm(dataset, index, form))}
+                removeDataTypeQueryForm={(index) => dispatch(removeDataTypeQueryForm(dataset, index))}
+            />
+            {hasIndividuals &&
+                !isFetchingSearchResults &&
+                (showTabs ? (
+                    <Tabs defaultActiveKey="1" onChange={onTabChange} activeKey={activeKey}>
```
v-rocheleau, line 101: Reference `TAB_KEYS.INDIVIDUAL` property for `defaultActiveKey` prop.

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -1,265 +1,123 @@
...
+                        <TabPane tab="Individual" key="1">
+                            <IndividualsTable data={searchResults.searchFormattedResults} />
+                        </TabPane>
+                        {hasBiosamples && (
+                            <TabPane tab="Biosamples" key="2">
```
v-rocheleau, line 106: Reference `TAB_KEYS.BIOSAMPLE` property for `key` prop.

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -1,265 +1,123 @@
...
+                                <BiosamplesTable data={searchResults.searchFormattedResultsBiosamples} />
+                            </TabPane>
+                        )}
+                        {hasExperiments && (
+                            <TabPane tab="Experiments" key="3">
```
v-rocheleau, line 111: Reference `TAB_KEYS.EXPERIMENT` property for `key` prop.

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -0,0 +1,180 @@
...
+    setSelectedRows,
+    performIndividualsDownloadCSVIfPossible,
+    performBiosamplesDownloadCSVIfPossible,
+    performExperimentsDownloadCSVIfPossible,
+} from "../../modules/explorer/actions";
+
+const ExplorerSearchResultsTable = ({ data, activeTab, ...props }) => {
+    const { dataset } = useParams();
+    const [currentPage, setCurrentPage] = useState(1);
+    const [pageSize] = useState(25);
```
v-rocheleau, line 20: If we dont update the `pageSize` state we can move it to a const outside the component

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -0,0 +1,180 @@
...
+        if (activeTab === "experiments") {
+            return dispatch(performExperimentsDownloadCSVIfPossible(dataset, ...args));
+        }
+    };
+
+    const onPageChange = (pageObj) => {
+        setCurrentPage(pageObj.current);
+    };
+
+    const tableStyle = {
```
v-rocheleau, line 56: move to const outside of component

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -0,0 +1,180 @@
...
+    performIndividualsDownloadCSVIfPossible,
+    performBiosamplesDownloadCSVIfPossible,
+    performExperimentsDownloadCSVIfPossible,
+} from "../../modules/explorer/actions";
+
+const ExplorerSearchResultsTable = ({ data, activeTab, ...props }) => {
+    const { dataset } = useParams();
+    const [currentPage, setCurrentPage] = useState(1);
+    const [pageSize] = useState(25);
+    const [numResults, setNumResults] = useState(0);
```
v-rocheleau, line 21: Not `useState(data.length)` ?

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -0,0 +1,180 @@
...
+    };
+
+    const tableStyle = {
+        opacity: fetchingSearch ? 0.5 : 1,
+        pointerEvents: fetchingSearch ? "none" : "auto",
+    };
+
+    useEffect(() => {
+        setNumResults(data.length);
+    }, [currentPage, pageSize]);
```
v-rocheleau, line 63: I think we can remove this useEffect, the effect function only depends on the `data` prop.

src/components/explorer/IndividualsTable.js
```diff
@@ -0,0 +1,77 @@
...
+    return (
+        <>
+            <Link
+                to={(location) => ({
+                    pathname: withBasePath(`data/explorer/individuals/${individual.id}/overview`),
+                    state: { backUrl: location.pathname },
+                })}
+            >
+                {individual.id}
+            </Link>{" "}
```
v-rocheleau, line 19: What is `{" "}` for?

src/modules/explorer/reducers.js
```diff
@@ -122,7 +126,44 @@ export const explorer = (
...
+            return {
+                ...state,
+                isFetchingDownload: false,
+            };
+        case PERFORM_BIOSAMPLE_CSV_DOWNLOAD.FINISH:
+            return {
+                ...state,
+                isFetchingDownload: false,
+            };
+        // ---
```
v-rocheleau, line 147: rm

src/modules/explorer/reducers.js
```diff
@@ -122,7 +126,44 @@ export const explorer = (
...
+            return {
+                ...state,
+                isFetchingDownload: false,
+            };
+        case PERFORM_EXPERIMENT_CSV_DOWNLOAD.FINISH:
+            return {
+                ...state,
+                isFetchingDownload: false,
+            };
+        // ---
```
v-rocheleau, line 166: rm

v-rocheleau (changes requested):

src/components/explorer/IndividualsTable.js
davidlougheed: forces a space after a closing tag because of the way react handles spaces in JSX

src/components/explorer/BiosamplesTable.js
```diff
@@ -0,0 +1,200 @@
...
+    biosample: PropTypes.string.isRequired,
+    alternateIds: PropTypes.arrayOf(PropTypes.string),
+    individualId: PropTypes.string.isRequired,
+};
+
+const customPluralForms = {
+    Serology: "Serologies",
+};
+
+const pluralize = (word, count) => {
```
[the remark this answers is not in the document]
noctillion: Absolutely, great suggestion and definitely worth considering, but the current implementation fulfills the current needs, however if you anticipate pluralization needs to become more complex in the future I can use the library without problem.. please let me know

src/components/explorer/ExplorerDatasetSearch.js
noctillion: Thanks for the suggestion, however here the code is ensuring there is at least one exp/biosample available before setting hasExp/Biosample to true. This avoid to display tables without data

v-rocheleau: Sorry I was not clear. What I meant is that I think it can be written like so:
```javascript
const hasBiosamples =
        searchResults &&
        searchResults.searchFormattedResultsBiosamples &&
        searchResults.searchFormattedResultsBiosamples.length;
```
Which should still evaluate to `false` when there is no data.

An even safer check would be:
```javascript
const hasBiosamples =
        searchResults &&
        Array.isArray(searchResults.searchFormattedResultsBiosamples) &&
        searchResults.searchFormattedResultsBiosamples.length;
```
This way we make sure `searchResults.searchFormattedResultsBiosamples` is an array, and that its `length` property is truthy (above 0).

Very minor suggestion, the original should be fine, but in javascript I prefer to be on the safer side with types.

src/components/explorer/ExplorerSearchResultsTable.js
noctillion: I am not sure here why would be better move it out? But if it is advisable to move it please let me know!

src/components/explorer/ExplorerSearchResultsTable.js
noctillion: Perfect!

src/components/explorer/BiosamplesTable.js
noctillion: Good idea! Moved!

src/components/explorer/ExplorerDatasetSearch.js
noctillion: I see!! Thank you very much for the clarification, you are right, the safer option is better!

src/components/explorer/ExplorerSearchResultsTable.js
v-rocheleau: Nvm, I am blind and didnt notice the ternary. All good as it is

src/components/explorer/ExplorerDatasetSearch.js
v-rocheleau: good call with the `hasNonEmptyArrayProperty`

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -59,27 +62,20 @@ const ExplorerDatasetSearch = () => {
     };
 
     const resetPageNumber = () => {
```
v-rocheleau, line 64: Function name no longer describes its purpose, could be renamed to `performSearch`

v-rocheleau (changes requested):
very minor request, will approve after

src/components/explorer/ExplorerDatasetSearch.js
noctillion: Yes, better name. Changed!

v-rocheleau (approved):
lgtm

PR #233, opened 2023-05-04
[12] Use auth code + PKCE flow for web authn 

davidlougheed:
to review this, you must be using the latest feat/auth-service branch of the main bento repo


src/lib/auth/performAuth.js
```diff
@@ -0,0 +1,120 @@
...
+        const paramState = params.get("state");
+        if (localState !== paramState) {
+            message.error("Error encountered during sign-in: state mismatch");
+            console.error("state mismatch");
+            setLSNotSignedIn();
+            return;
+        }
+
+        const verifier = localStorage.getItem(PKCE_LS_VERIFIER);
+        localStorage.removeItem(PKCE_LS_VERIFIER);
```
v-rocheleau, line 113: Minor. We could wrap the few repetitions of `localStorage.getItem(key) ... localStorate.removeItem(key)` in a utils function named something like `popLsItem`

src/components/App.js
```diff
@@ -35,183 +38,235 @@ const NotificationsContent = lazy(() => import("./notifications/NotificationsCon
 
 const SIGN_IN_WINDOW_FEATURES = "scrollbars=no, toolbar=no, menubar=no, width=800, height=600";
 
-class App extends Component {
-    constructor(props) {
-        super(props);
+const CALLBACK_PATH = withBasePath("callback");
+const CallbackContent = () => <SitePageLoading />;
```
v-rocheleau, line 42: Why this additional wrapping?

v-rocheleau (changes requested):
First pass, looking good so far! Will go more into components and redux stuff tomorrow.

src/lib/auth/performAuth.js
davidlougheed: good idea

src/lib/auth/utils.js
```diff
@@ -0,0 +1,6 @@
+// noinspection JSCheckFunctionSignatures
+export const buildUrlEncodedData = obj =>
+    Object.entries(obj).reduce((params, [k, v]) => params.set(k, v.toString()) || params, new URLSearchParams());
```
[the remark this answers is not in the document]
davidlougheed: it's a hack using or-shortcutting - an expression-only equivalent to
```js
(params, [k, v]) => { params.set(k, v.toString()); return params; }
```
since params.set returns undefined. it's unclear though, i'll change these instances.

src/session.worker.js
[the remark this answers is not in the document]
davidlougheed: the actual talking to the IdP used to happen via the gateway (thus the ping), but now the main thread does it in a redux action.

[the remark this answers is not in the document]
v-rocheleau: I see, thanks for the explanation

src/components/App.js
```diff
@@ -35,183 +38,235 @@ const NotificationsContent = lazy(() => import("./notifications/NotificationsCon
...
+    const openIdConfig = useSelector(state => state.openIdConfiguration.data);
 
-            // Don't bother trying to create the event relay connection if the user isn't authenticated
-            if (!this.props.user) return null;
+    const [lastIsAuthenticated, setLastIsAuthenticated] = useState(false);
+
+    const [didPostLoadEffects, setDidPostLoadEffects] = useState(false);
+
+    // TODO: add in localstorage flag to condition to avoid false positives
+    const isInAuthPopup = !!window.opener;
```
v-rocheleau, line 82: False positive when coming in from public
image

v-rocheleau (approved):
Behaves as expected. Did some testing in keycloak admin with session revocation, 2 min session check works. Authz working as expected.

PR #234, opened 2023-05-05
[12] Fix (#1623): drop box json viewer expand/collapse glitch

v-rocheleau:
**Bug:** The [package](https://github.com/mac-s-g/react-json-view) (react-json-view) we use for json display has not been updated in 2 years, and there are a number of open issues on their repo describing the exact same bug with large json documents.

Tried using a maintained fork of react-json-view called [json-viewer](https://github.com/TexteaInc/json-viewer), but ran into too many compat issues with react and antd major versions updates. We will be able to make the switch once we are on react 17.x and antd 4.x.

**Fix:** Instead of relying on the buggy react-json-view collapsed array selector, I made wrapper components (JsonArrayDisplay, JsonObjectDisplay) that take control of grouped items selection using antd.


src/components/manager/ManagerDropBoxContent.js
```diff
@@ -143,7 +143,8 @@ const FileDisplay = ({file, tree, treeLoading}) => {
                 setLoadingFileContents(true);
                 const r = await fetch(urisByFilePath[file]);
                 if (r.ok) {
-                    setFileContents({...fileContents, [file]: await r.text()});
+                    const parsed = JSON.parse(await r.text());
```
davidlougheed, line 146: this doesn't work if the file isn't JSON - the viewer supports other files (READMEs, text files, etc.) too!

davidlougheed (changes requested):

src/components/manager/ManagerDropBoxContent.js
v-rocheleau: good point, this is taken into account now

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -194,15 +187,11 @@ const FileDisplay = ({file, tree, treeLoading}) => {
...
-                    <ReactJson
-                        src={JSON.parse(fileContents[file] || "{}")}
-                        displayDataTypes={false}
-                        enableClipboard={false}
-                        name={null}
-                        collapsed={true}
-                    />
-                );
+                if (Array.isArray(fileContents[file])) {
+                    return ( <JsonArrayDisplay doc={fileContents[file] || []} standalone/> );
```
davidlougheed, line 191: formatting: inconsistent spacing + needless parens with this line and the return below it

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -121,17 +119,7 @@ const FileDisplay = ({file, tree, treeLoading}) => {
...
-            if (fileExt === "pdf") {
-                setLoadingFileContents(true);
-            }
-
-            if (!textFormat) return;
-
-            if (fileContents.hasOwnProperty(file)) {
-                return;
-            }
+            if (!file || !textFormat || fileContents.hasOwnProperty(file)) return;
```
davidlougheed, line 122: this removes some special pdf handling code

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -39,6 +37,7 @@ import TableSelectionModal from "./TableSelectionModal";
 import {STEP_INPUT} from "./workflowCommon";
 import {withBasePath} from "../../utils/url";
 import {dropBoxTreeStateToPropsMixinPropTypes, workflowsStateToPropsMixin} from "../../propTypes";
+import { JsonArrayDisplay, JsonObjectDisplay } from "../JsonDisplay";
```
davidlougheed, line 40: move this up to near table selection modal to be grouped with the other component import

davidlougheed (changes requested):

src/components/manager/ManagerDropBoxContent.js
v-rocheleau: I think the removed `setLoadingFileContents(true)` was redundant since it is called in the try block also. So if `fileExt === "pdf"` is true it would be called two times. I didnt notice any change visually either way so I factored it out.

davidlougheed: pdf doesn't reach the try/catch since it isn't a text-formatted

v-rocheleau: Oh right, fixed it

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -194,15 +193,10 @@ const FileDisplay = ({file, tree, treeLoading}) => {
...
-                        displayDataTypes={false}
-                        enableClipboard={false}
-                        name={null}
-                        collapsed={true}
-                    />
-                );
+                if (Array.isArray(fileContents[file])) {
+                    return (<JsonArrayDisplay doc={fileContents[file] || []} standalone/>);
+                }
+                return (<JsonObjectDisplay doc={fileContents[file] || {}}/>);
```
davidlougheed, line 199: this being the default case means that the title "JSON object" shows when it is loading. I think it would be better if it showed nothing while loading.

also this will break on JSON primitives - so if the file just has a string or number in it. while this isn't hugely important for all drop box files we see, I think for completeness it would be good to check if it is 

* an array
* another form of object
* else: just render in a plain ReactJson (for primitives)

while loading, you can just render a `<div />`.

davidlougheed (changes requested):

src/components/manager/ManagerDropBoxContent.js
v-rocheleau: Now returning a div when loading.
Also wrapped JsonObjectDisplay and JsonArrayDisplay inside of JsonDisplay for usability. Fixed primitive behaviour in JsonPropertyDisplay.

gsfk (commented):
I tried opening a json with this structure: 
```
{
  experiments: {...},
  resources: {...}
}
```
and got the revenge of "JSON object" (along with some caps, not sure the caps matter):

Screenshot from 2023-05-17 14-36-11

davidlougheed (approved):
lgtm

PR #235, opened 2023-05-16
Features/data type counts

noctillion:
Add filtering for queryable and non-zero count dataTypes.

In this PR, the DiscoveryQueryBuilder and DataTypeExplorationModal components are updated to filter out the data types that are not queryable or have zero counts.

Changes include:

- Filtering non-queryable dataTypes: The component now lists data types with the queryable property set to true.
- Filtering zero count dataTypes: The component also filters out the data types with zero counts.

### DIFF
--- a/src/components/discovery/DataTypeExplorationModal.js
+++ b/src/components/discovery/DataTypeExplorationModal.js
@@ -77,22 +77,26 @@ class DataTypeExplorationModal extends Component {
                 </Radio.Group>
                 <Tabs>
                     {Object.values(this.props.dataTypes).flatMap(ds => (ds.items ?? [])
-                        .filter(dataType => dataType.queryable && dataType.count > 0)
+                        .filter(dataType => (dataType.queryable ?? true) && dataType.count > 0)
                         .map(dataType =>
-                        <Tabs.TabPane tab={dataType.label ?? dataType.id} key={dataType.id}>
-                            {this.state.view === "tree" ? (
-                                <SchemaTree schema={dataType.schema} />
-                            ) : (
-                                <>
-                                    <Input.Search allowClear={true}
-                                                  onChange={e => this.onFilterChange(e.target.value)}
-                                                  placeholder="Search for a field..." style={{marginBottom: "16px"}} />
-                                    <Table bordered={true}
-                                           columns={FIELD_COLUMNS}
-                                           dataSource={this.getTableData(dataType)} />
-                                </>
-                            )}
-                        </Tabs.TabPane>
+                            <Tabs.TabPane tab={dataType.label ?? dataType.id} key={dataType.id}>
+                                {this.state.view === "tree" ? (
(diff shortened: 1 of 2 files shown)


src/components/discovery/DataTypeExplorationModal.js
```diff
@@ -76,7 +76,9 @@ class DataTypeExplorationModal extends Component {
                     <Radio.Button value="table"><Icon type="table" /> Table Detail View</Radio.Button>
                 </Radio.Group>
                 <Tabs>
-                    {Object.values(this.props.dataTypes).flatMap(ds => (ds.items ?? []).map(dataType =>
+                    {Object.values(this.props.dataTypes).flatMap(ds => (ds.items ?? [])
+                        .filter(dataType => dataType.queryable && dataType.count > 0)
```
davidlougheed, line 80: this is good for now, but in the future we will need to filter by data type at the level of the dataset, not the whole of the instance. maybe that can be a separate ticket?

davidlougheed: also, if queryable isn't set, it should default to true

so this would look like

```js
.filter(dataType => (dataType.queryable ?? true) && dataType.count > 0)
```

src/components/discovery/DataTypeExplorationModal.js
```diff
@@ -76,7 +76,9 @@ class DataTypeExplorationModal extends Component {
...
+                        .map(dataType =>
                         <Tabs.TabPane tab={dataType.label ?? dataType.id} key={dataType.id}>
```
davidlougheed, line 82: indent this block of JSX

src/components/discovery/DiscoveryQueryBuilder.js
```diff
@@ -137,9 +137,12 @@ class DiscoveryQueryBuilder extends Component {
                 {this.props.servicesInfo
                     .filter(s => (this.props.dataTypes[s.id]?.items ?? []).length)
                     .flatMap(s =>
-                        this.props.dataTypes[s.id].items.map(dt =>
-                            <Menu.Item key={`${s.id}:${dt.id}`}>{dt.label ?? dt.id}</Menu.Item>
-                        ))
+                        this.props.dataTypes[s.id].items
+                            .filter(dt => dt.queryable && dt.count > 0)
```
davidlougheed, line 141: same here, use (dt.queryable ?? true) to make it default to true if undefined

davidlougheed (changes requested):
minor changes but otherwise looks good for v12!

davidlougheed (approved):
lgtm

src/components/discovery/DataTypeExplorationModal.js
noctillion: Let's handle this in a separate ticket. As for the default value of queryable, you're absolutely right. I've updated the code to treat queryable as true when it's not set. Thanks for your insights!!

PR #241, opened 2023-05-25
[12.1] refact: remove withBasePath

davidlougheed:
all instances should be required to have their own subdomain anyway, so this is a lot of extra needless code that dates back to the CHORD days.


brouillette (commented):
👌

PR #242, opened 2023-05-29
[12.1]: feat: add loading indicator to notification 'mark as read' button

davidlougheed:
also rewrites `NotificationList` as a functional component


gsfk (approved):
Appears to work as intended. 

Only comment, arguably outside the scope of this PR, is that you can only click through to the run details from an old notification, and not on the "live" notification itself when it first appears, which seems odd.

PR #244, opened 2023-05-30
[12] fix: fetch Gohan overview

brouillette:
Obtains the gohan url from service artifacts rather than services

### DIFF
--- a/src/modules/explorer/actions.js
+++ b/src/modules/explorer/actions.js
@@ -154,7 +154,7 @@ export const setIgvPosition = (igvPosition) => ({
 });
 
 export const performGetGohanVariantsOverviewIfPossible = () => (dispatch, getState) => {
-    const gohanUrl = getState()?.services?.gohan?.url;
+    const gohanUrl = getState()?.services?.itemsByArtifact?.gohan?.url;
     if (!gohanUrl) return;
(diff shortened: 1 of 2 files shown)


src/modules/explorer/actions.js
```diff
@@ -154,7 +154,7 @@ export const setIgvPosition = (igvPosition) => ({
 });
 
 export const performGetGohanVariantsOverviewIfPossible = () => (dispatch, getState) => {
-    const gohanUrl = getState()?.services?.gohan?.url;
+    const gohanUrl = getState()?.services?.itemsByArtifact?.gohan?.url;
```
davidlougheed, line 157: itemsByArtifact is deprecated; use itemsByKind

davidlougheed (changes requested):

v-rocheleau (dismissed):
Works locally for me, but there is another gohan action we probably want to change too in [discovery](https://github.com/bento-platform/bento_web/blob/ff5f2bee8cbc30cab8cf729ee0af936f561d9ff8/src/modules/discovery/actions.js#L9)

davidlougheed (approved):
lgtm

v-rocheleau (approved):

PR #245, opened 2023-05-31
[12] display table header when ind 0

noctillion:
This PR makes changes to allow the display table header when search results are 0


v-rocheleau (changes requested):

v-rocheleau (approved):
lgtm

PR #248, opened 2023-06-01
patch: count keys instead of adding values

brouillette:
obtains the number of variant files by counting the number of sampleIds present rather than summing the values


v-rocheleau (approved):
Lgtm, tested locally.

PR #249, opened 2023-06-06
[12.1] feat: drop box permissions awareness + delete fn

davidlougheed:
requires drop box pr-28 image and main bento repo feat/new-drop-box


src/lib/auth/permissions.js
```diff
@@ -0,0 +1,3 @@
+export const viewDropBox = "view:drop_box";
+export const ingestDropBox = "ingest:drop_box";
```
v-rocheleau, line 2: Looking at how it is being used, I would rename this permission to use the "upload" verb instead of "ingest", given that we already use the term "ingestion" for dataset tables

src/lib/auth/utils.js
```diff
@@ -10,3 +14,27 @@ export const getIsAuthenticated = idTokenContents =>
     !!idTokenContents && Math.round((new Date()).getTime() / 1000) < idTokenContents.exp;
 
 export const makeAuthorizationHeader = token => token ? {"Authorization": `Bearer ${token}`} : {};
+
+// TODO: move hooks to own file
+
+export const useAuthorizationHeader = () => {
```
v-rocheleau, line 20: Neat

src/lib/auth/utils.js
```diff
@@ -10,3 +14,27 @@ export const getIsAuthenticated = idTokenContents =>
...
+    const {accessToken} = useSelector(state => state.auth);
+    return useMemo(
+        () => accessToken => accessToken ? {"Authorization": `Bearer ${accessToken}`} : {},
+        [accessToken]);
+};
+
+export const useResourcePermissions = (resource) => {
```
v-rocheleau, line 27: Nice addition for authz! 
Minor suggestion (could be a todo comment): It might be worthwhile to extend this hook to a `hasResourcePermissions => (resource, permission)` that evaluates if the permission is valid or not.

v-rocheleau (changes requested):
Small suggestions

v-rocheleau (approved):
Lgtm. We'll need to fix the download button for the authz

v-rocheleau (approved):
Lgtm & working as expected

PR #250, opened 2023-06-06
[12.1] feat: user profile page

davidlougheed:
Allows users to view their issuer + subject ID. The page is available from the user dropdown in the nav bar.


gsfk (approved):
Works well. Issuer and subject id now much easier to find!

I didn't see a way to test failure mode but I don't see any issues.

PR #253, opened 2023-06-08
date and time of last ingestion

noctillion:
This PR includes changes to additionally to the date of the latest ingestion display, the time

### DIFF
--- a/src/components/manager/runs/RunLastContent.js
+++ b/src/components/manager/runs/RunLastContent.js
@@ -4,7 +4,12 @@ import {useSelector} from "react-redux";
 import PropTypes from "prop-types";
 
 const COLUMNS_LAST_CONTENT = [
-    {title: "Date", dataIndex: "date", key: "date"},
+    {
+        title: "Date",
+        dataIndex: "date",
+        key: "date",
+        render: (date) => formatDate(date),
+    },
     {title: "Data Type", dataIndex: "dataType", key: "dataType"},
     {title: "Table ID", dataIndex: "tableId", key: "tableId"},
     {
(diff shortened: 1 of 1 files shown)


src/components/manager/runs/RunLastContent.js
```diff
@@ -113,13 +124,13 @@ const processIngestions = (data, currentTables) => {
                 : [fileNameFromPath(filePaths)];
 
             const dateStr = getDateFromEndTime(run.details.run_log.end_time);
-            const date = Date.parse(dateStr);
+            const date = Date.parse(run.details.run_log.end_time);
 
-            const currentIngestion = { date: dateStr, dataType, tableId, fileNames };
+            const currentIngestion = { date: dateStr, dataType, tableId, fileNames, newdate: date };
```
davidlougheed, line 129: `newdate` is not a very informative property name I think - what does it do differently vs dateStr?

davidlougheed (changes requested):

src/components/manager/runs/RunLastContent.js
```diff
@@ -112,16 +129,15 @@ const processIngestions = (data, currentTables) => {
                 ? filePaths.map(fileNameFromPath)
                 : [fileNameFromPath(filePaths)];
 
-            const dateStr = getDateFromEndTime(run.details.run_log.end_time);
-            const date = Date.parse(dateStr);
+            const date = Date.parse(run.details.run_log.end_time);
 
-            const currentIngestion = { date: dateStr, dataType, tableId, fileNames };
+            const currentIngestion = { date: date, dataType, tableId, fileNames };
```
davidlougheed, line 134: this can just be `{ date, dataType, ...}` to match the other short forms here
other than that, looks good to me

davidlougheed (changes requested):

src/components/manager/runs/RunLastContent.js
noctillion: Sure, Thanks!

davidlougheed (approved):
lgtm!

PR #254, opened 2023-06-09
feat: proper sign out functionality re-do

davidlougheed:


v-rocheleau (approved):
Lgtm, we are going with session termination after all?

davidlougheed:
This should only terminate the session for this particular client, which I think works fine - turns out I was misunderstanding a bunch of OIDC stuff (again)

PR #262, opened 2023-06-20
Fix/project creation modal

v-rocheleau:
**Issue:** Conditional rendering in ManagerProjectDatasetContent would not render a ProjectCreationModal if `projectMenuItems.length === 0`. This prevented the modal to appear when clicking on the "Create Project" button.

**Solution:** Make `ManagerProjectDatasetContent` render a ProjectCreationModal, so that it can react to redux state changes.

**Additional change:** refactored ProjectCreationModal to a functional component.


davidlougheed (approved):
oops, ty for the fix

PR #263, opened 2023-06-20
feat: default linked-field-set button

v-rocheleau:
Added new button to quickly create a linked field set for biosamples, experiments and phenopackets.


gsfk (commented):
Works as expected, although it offers to create a second linked field set when one already exists, perhaps @davidlougheed can let us know if this is an error or not.

davidlougheed:
maybe disable if one exists yeah

davidlougheed (approved):
lgtm

PR #266, opened 2023-07-03
feat: fetch extra_properties types dynamically from Katsu

noctillion:
Extra property schema types fetching and integration

This PR introduces the fetching and integration of extra property schema types into the form selection options.
Related to Feature #1683

### DIFF
--- a/src/components/manager/projects/ProjectJsonSchemaForm.js
+++ b/src/components/manager/projects/ProjectJsonSchemaForm.js
@@ -15,7 +15,7 @@ const ajv = new Ajv({
 const validateSchema = ajv.getSchema("http://json-schema.org/draft-07/schema");
 
 const getSchemaTypeOptions = (schemaTypes) => {
-    if (schemaTypes instanceof Object) {
+    if (typeof schemaTypes === "object" && schemaTypes !== null) {
         return Object.entries(schemaTypes).map(([key, value]) => ({
             key,
             value: key,
--- a/src/components/manager/projects/ProjectJsonSchemaForm.js
+++ b/src/components/manager/projects/ProjectJsonSchemaForm.js
@@ -14,6 +14,18 @@ const ajv = new Ajv({
 // Does not actually query over http, the URI is the key to the draft-07 meta-schema
 const validateSchema = ajv.getSchema("http://json-schema.org/draft-07/schema");
 
+const getSchemaTypeOptions = (schemaTypes) => {
+    if (schemaTypes instanceof Object) {
+        return Object.entries(schemaTypes).map(([key, value]) => ({
+            key,
+            value: key,
+            text: value.toUpperCase(),
+        }));
+    } else {
+        return [];
+    }
+};
+
 const ProjectJsonSchemaForm = ({ style, schemaTypes, initialValues, setFileContent, fileContent, form }) => {
 
     const onDrop = useCallback((files) => {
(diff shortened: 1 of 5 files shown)


src/modules/metadata/reducers.js
```diff
@@ -9,6 +9,7 @@ import {
     DELETE_PROJECT,
     SAVE_PROJECT,
 
+    FETCH_EXTRA_PROPERTIES_SCHEMA_TYPES,
```
davidlougheed, line 12: move this to between `FETCH_OVERVIEW_SUMMARY` and `CREATE_PROJECT_JSON_SCHEMA`

src/components/manager/projects/ProjectJsonSchemaForm.js
```diff
@@ -14,6 +14,18 @@ const ajv = new Ajv({
...
 
+const getSchemaTypeOptions = (schemaTypes) => {
+    if (schemaTypes instanceof Object) {
+        return Object.entries(schemaTypes).map(([key, value]) => ({
+            key,
+            value: key,
+            text: value.toUpperCase(),
+        }));
+    } else {
+        return [];
```
davidlougheed, line 25: when is this not an object?

src/modules/metadata/reducers.js
```diff
@@ -42,6 +43,7 @@ export const projects = (
         isSavingDataset: false,
         isDeletingDataset: false,
 
+        extraPropertiesSchemaTypes: {},
```
davidlougheed, line 46: add newline below;
missing `isFetchingExtraPropertiesSchemaInfo: false` initialization here

davidlougheed (changes requested):
overall looking good! minor requests

src/components/manager/projects/ProjectJsonSchemaModal.js
```diff
@@ -67,6 +71,7 @@ ProjectJsonSchemaModal.propTypes = {
 
     onOk: PropTypes.func,
     onCancel: PropTypes.func,
+    extraPropertiesSchemaTypes: PropTypes.object.isRequired,
```
v-rocheleau, line 74: I think this is not needed since `extraPropertiesSchemaTypes` is obtained from redux with `useSelector`

v-rocheleau (changes requested):
Looking good, small request

v-rocheleau (approved):
Lgtm, but would like an explanation for this [comment](https://github.com/bento-platform/bento_web/pull/266/files#r1268530326) too

src/components/manager/projects/ProjectJsonSchemaForm.js
noctillion: I changed instanceof to typeof === "object". The previous way could cause problems in some situations, but this new way makes sure that schemaTypes is always an object and checks for no null. It should be more reliable. Thanks for pointing that out!

noctillion:
> Lgtm, but would like an explanation for this [comment](https://github.com/bento-platform/bento_web/pull/266/files#r1268530326) too

Sure thing!! Thanks for your review and insights!!

src/modules/metadata/reducers.js
```diff
@@ -42,6 +44,8 @@ export const projects = (
...
+        isFetchingExtraPropertiesSchemaInfo: false,
```
davidlougheed, line 48: rename this variable to `isFetchingExtraPropertiesSchemaTypes`

davidlougheed (changes requested):
very minor tweaks, otherwise looks good!

davidlougheed (approved):
lgtm

PR #267, opened 2023-07-04
 Enhance error handling in network fetch functions

SanjeevLakhwani:
This PR improves the error handling logic in the _unpaginatedNetworkFetch and _paginatedNetworkFetch functions, which are used to make API requests. Previously, when the server returned an error status, these functions would simply throw the HTTP status code and text. Now, they first attempt to parse the server's response and extract an error message from it. This allows the application to display more specific error messages to the user, providing a better user experience and clearer debugging information.

This change assumes that the server includes error messages in the body of error responses. If an error message from the server is not available, the HTTP status code and text are used as a fallback.

Changes:

- Update _unpaginatedNetworkFetch and _paginatedNetworkFetch functions to extract error messages from server responses.

Testing:

Please ensure that error handling still works as expected, both for errors where the server includes a message and for other errors. Please also ensure that other parts of the application that use these functions are still functioning correctly.

### DIFF
--- a/src/utils/actions.js
+++ b/src/utils/actions.js
@@ -1,29 +1,29 @@
 import fetch from "cross-fetch";
 
-import {message} from "antd";
+import { message } from "antd";
 
-import {BENTO_PUBLIC_URL, BENTO_URL, IDP_BASE_URL} from "../config";
+import { BENTO_PUBLIC_URL, BENTO_URL, IDP_BASE_URL } from "../config";
 
-export const basicAction = t => () => ({type: t});
+export const basicAction = (t) => () => ({ type: t });
 
-export const createNetworkActionTypes = name => ({
+export const createNetworkActionTypes = (name) => ({
     REQUEST: `${name}.REQUEST`,
     RECEIVE: `${name}.RECEIVE`,
     ERROR: `${name}.ERROR`,
     FINISH: `${name}.FINISH`,
 });
 
-export const createFlowActionTypes = name => ({
+export const createFlowActionTypes = (name) => ({
     BEGIN: `${name}.BEGIN`,
     END: `${name}.END`,
     TERMINATE: `${name}.TERMINATE`,
 });
 
-
 const _unpaginatedNetworkFetch = async (url, _baseUrl, req, parse) => {
     const response = await fetch(url, req);
     if (!response.ok) {
-        throw `${response.status} ${response.statusText}`;
+        const errorData = await parse(response);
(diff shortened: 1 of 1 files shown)


src/utils/actions.js
```diff
@@ -1,29 +1,29 @@
 import fetch from "cross-fetch";
 
-import {message} from "antd";
+import { message } from "antd";
 
-import {BENTO_PUBLIC_URL, BENTO_URL, IDP_BASE_URL} from "../config";
+import { BENTO_PUBLIC_URL, BENTO_URL, IDP_BASE_URL } from "../config";
 
-export const basicAction = t => () => ({type: t});
+export const basicAction = (t) => () => ({ type: t });
```
davidlougheed, line 7: why are you applying prettier to bento web? it makes the diffs very hard to read.

davidlougheed: i'll just do a whole-pass prettier run after this; i'd rather have readable PR diffs than blame.

src/utils/actions.js
```diff
@@ -33,13 +33,14 @@ const _paginatedNetworkFetch = async (url, baseUrl, req, parse) => {
     const _fetchNext = async (pageUrl) => {
         const response = await fetch(pageUrl, req);
         if (!response.ok) {
-            throw "Invalid response encountered";
+            const errorData = await parse(response);
```
davidlougheed, line 36: what happens if parsing fails? i.e. not a JSON response - i think this would result in the JSON parse error being the message instead of 'Invalid response encountered' as desired

src/utils/actions.js
```diff
@@ -48,65 +49,72 @@ const _paginatedNetworkFetch = async (url, baseUrl, req, parse) => {
...
+                );
+                dispatch({
+                    type: types.RECEIVE,
+                    ...params,
+                    ...(data === null ? {} : { data }),
+                    receivedAt: Date.now(),
+                });
+                if (onSuccess) await onSuccess(data);
+            } catch (e) {
+                const errorMsg = e.message || err;
```
davidlougheed, line 103: the `err` parameter is used to add nice errors in actions, and this is confusing behaviour if it gets overwritten in some cases - maybe we should remove it completely, or use it as a prefix like `${err}: ${e.message}` if `err` is set

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

v-rocheleau (approved):

PR #269, opened 2023-07-05
fix: IGV.js crash

davidlougheed:


davidlougheed:
(deployed on qa)

gsfk (approved):
Looks good, working well on QA. 

I'm curious if the more explicit import fixes the problem without pinning the version, but unable to test locally at the moment.

PR #281, opened 2023-08-15
[13] fix: retain search table settings and order

noctillion:
This PR implements changes to retain the sorting and pagination information in the explorer results tables. These changes allow the navigation back to the tables, preserving the user's preferred sorting order and the selected tab and pagination. The main changes introduced with this PR are:

Persistent sorting: The sorting order for the columns is now saved and reapplied when navigating back to the explorer result tables

Saved pagination state: The current page is stored and retrieved when returning to the tables, keeping the user's position in the dataset.

Redux integration: Updated the redux actions and state to handle the storing and retrieval of sorting and pagination information, making it accessible across different components.

Code refactoring: Refactored the code including the implementation of custom sorting hook and the ExplorerSearchResultsTable component.

### DIFF
--- a/src/components/explorer/ExplorerDatasetSearch.js
+++ b/src/components/explorer/ExplorerDatasetSearch.js
@@ -15,6 +15,7 @@ import {
     removeDataTypeQueryForm,
     updateDataTypeQueryForm,
     setSelectedRows,
+    resetTableSortOrder,
 } from "../../modules/explorer/actions";
 
 import IndividualsTable from "./IndividualsTable";
@@ -68,6 +69,7 @@ const ExplorerDatasetSearch = () => {
     };
 
     const performSearch = () => {
+        dispatch(resetTableSortOrder(dataset));
         dispatch(performSearchIfPossible(dataset));
     };
 
--- a/src/components/explorer/ExplorerDatasetSearch.js
+++ b/src/components/explorer/ExplorerDatasetSearch.js
@@ -102,16 +102,22 @@ const ExplorerDatasetSearch = () => {
                 (showTabs ? (
                     <Tabs defaultActiveKey={TAB_KEYS.INDIVIDUAL} onChange={onTabChange} activeKey={activeKey}>
                         <TabPane tab="Individual" key={TAB_KEYS.INDIVIDUAL}>
-                            <IndividualsTable data={searchResults.searchFormattedResults} />
+                            <IndividualsTable
+                                data={searchResults.searchFormattedResults}
+                                datasetID={dataset}/>
                         </TabPane>
                         {hasBiosamples && (
                             <TabPane tab="Biosamples" key={TAB_KEYS.BIOSAMPLES}>
-                                <BiosamplesTable data={searchResults.searchFormattedResultsBiosamples} />
+                                <BiosamplesTable
+                                    data={searchResults.searchFormattedResultsBiosamples}
+                                    datasetID={dataset}/>
                             </TabPane>
                         )}
                         {hasExperiments && (
                             <TabPane tab="Experiments" key={TAB_KEYS.EXPERIMENTS}>
-                                <ExperimentsTable data={searchResults.searchFormattedResultsExperiment} />
+                                <ExperimentsTable
+                                    data={searchResults.searchFormattedResultsExperiment}
+                                    datasetID={dataset}/>
                             </TabPane>
                         )}
                     </Tabs>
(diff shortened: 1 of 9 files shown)


src/components/explorer/ExplorerIndividualContent.js
```diff
@@ -42,6 +42,7 @@ class ExplorerIndividualContent extends Component {
 
         this.state = {
             backUrl: null,
+            currentTab: null,
             selectedTab: "overview",
         };
     }
```
gsfk, line 48: Is this an unrelated fix? "tab" here seems like it refers to the top level tabs : Overview, Exporer, etc rather than the tabs in the search results table. (also, is `selectedTab` dead code? If not, what's the difference betweeen currentTab and selectedTab?)

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -83,6 +86,13 @@ const ExplorerSearchResultsTable = ({ data, activeTab, ...props }) => {
         ],
     };
 
+    const sortedInfo = useMemo(() => {
+        return {
+            order: props.sortOrder,
+            columnKey: props.sortColumnKey,
+        };
+    }, [props.sortOrder, props.sortColumnKey]);
```
gsfk, line 94: odd mixed style here (all other props in this component are destructured, ie called by `someName`, rather than `props.someName`).. although in general the repo overall has wildly varying styles...

gsfk (approved):
Tested as much as I could and works great. I left a few inline comments, none of which are crucial.

src/components/explorer/hooks/explorerHooks.js
```diff
@@ -0,0 +1,31 @@
...
+
+export const useSortedColumns = (data, tableSortOrder, columnsDefinition) => {
+    const sortColumnKey = tableSortOrder?.sortColumnKey;
+    const sortOrder = tableSortOrder?.sortOrder;
+
+    const sortData = (dataToSort, sortKey, order) => {
+        const column = columnsDefinition.find((col) => col.dataIndex === sortKey);
+        if (column && column.sorter) {
+            return [...dataToSort].sort((a, b) => {
+                return order === "ascend" ? column.sorter(a, b) : -column.sorter(a, b);
```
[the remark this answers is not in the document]
noctillion: Yes, you are right. I relied on the simplicity of negating the result of the column.sorter(a, b) call for the descending order. But, given that there is not more than that, I agree that swapping the parameter order is a better choice! Thanks for the suggestion!

src/components/explorer/ExplorerSearchResultsTable.js
noctillion: Good catch! thanks

src/components/explorer/IndividualsTable.js
```diff
@@ -9,10 +11,10 @@ const IndividualRender = ({individual}) => {
     return (
         <>
             <Link
-                to={(location) => ({
+                to={{
                     pathname: `/data/explorer/individuals/${individual.id}/overview`,
                     state: { backUrl: location.pathname },
```
davidlougheed, line 16: location is now undefined here

davidlougheed (changes requested):

src/components/explorer/ExplorerIndividualContent.js
noctillion: Sure! I removed the legacy code.

src/components/explorer/ExplorerDatasetSearch.js
```diff
@@ -34,19 +36,21 @@ const hasNonEmptyArrayProperty = (targetObject, propertyKey) => {
...
     const { dataset } = useParams();
+    const dispatch = useDispatch();
 
     const datasetsByID = useSelector((state) =>
         Object.fromEntries(
             state.projects.items.flatMap((p) => p.datasets.map((d) => [d.identifier, { ...d, project: p.identifier }])),
         ),
     );
+
+    const activeKey = useSelector((state) => state.explorer.activeTabByDatasetID[dataset]) || TAB_KEYS.INDIVIDUAL;
```
davidlougheed, line 48: nice! good move to keep this in redux

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -12,13 +12,22 @@ import {
...
+    data,
+    activeTab,
+    columns,
+    currentPage: initialCurrentPage,
+    sortOrder,
+    sortColumnKey,
+}) => {
     const { dataset } = useParams();
-    const [currentPage, setCurrentPage] = useState(1);
+    //const [currentPage, setCurrentPage] = useState(1);
```
davidlougheed, line 29: rm

davidlougheed (changes requested):
overall looking really good! made a few minor requests

noctillion:
Thanks for your review and insights!!

davidlougheed (approved):
lgtm

PR #287, opened 2023-08-31
feat!: delete variants on project/dataset deletion; load data types/workflows from service-registry; rm hardcoding artifacts

v-rocheleau:
refact!: load data types + workflows from new service registry (1.2.0+) & clean up dependents

### DIFF
--- a/src/components/datasets/Dataset.js
+++ b/src/components/datasets/Dataset.js
@@ -13,8 +13,8 @@ import {
 } from "../../modules/metadata/actions";
 
 import {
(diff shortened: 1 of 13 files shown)


davidlougheed (commented):
in my opinion data types should be cleared before deletion, not after success - this way, if a clear fails, we still have the dataset definition and so we can try again. otherwise, we need to go into Elasticsearch to clear the variants (or Katsu Postgres for others).

Also, it would be better if this iterated through all data types and created promises to clear them in parallel - in the future datasets might not be part of Katsu, and so Katsu will need to know about dataset deletion.

PR #289, opened 2023-09-05
fix: dataset datatypes counts loaded with user dependent data

v-rocheleau:


davidlougheed (approved):
Lgtm. Make sure to bump version in package.json and run npm install to sync to package-lock before tagging.

PR #291, opened 2023-09-06
feat: refactor RunLog component to be func + display ANSI colors

davidlougheed:
some colour should show up in WES logs now. This is controlled by Cromwell using [ANSI escape codes](https://en.wikipedia.org/wiki/ANSI_escape_code), not us directly.


v-rocheleau (approved):
Lgtm, sorry I saw this one late

PR #292, opened 2023-09-06
fix: render loading skeleton when loading igv urls

davidlougheed:
refact: memoize some track values


gsfk (approved):
Thanks! Tested locally and on QA.

PR #293, opened 2023-09-07
feat(explorer): show individual biosamples in an expandable table with linkable URLs

davidlougheed:


gsfk (commented):
okay, hmmm, the current way biosamples are shown is less than ideal, but is the best solution dozens of tabs?

so_many_tabs

gsfk (commented):
Yeah, better now. I like how the tags still appear even when collapsed. 

Minor issues / discussion: 

- Biosamples are still not clickable from "Experiments" results table 
- This introduces a second style of collapsible display, very similar to the one used in the Experiments tab. Shouldn't we just pick one? 
- somewhat outside the scope of this pr: why do we even bother listing biosample ids in the "Individual" results table. Experiments only get a count in the same table, and it can just look like a long meaningless list.

gsfk:
Yes, I think so, the new display looks better.

davidlougheed:
okay, I created a ticket for doing this for experiments: https://206.12.92.46/issues/1894

Regarding your other question, it shows the list of biosamples because the Individuals display used to be the only table available and I think we figured people may want to jump to specific biosamples easily - definitely worth rethinking

PR #298, opened 2023-09-11
feat(explorer): individual phenopackets JSON view

davidlougheed:
adds a tab to the individual view showing phenopackets JSON for that individual. as requested by David B.

the download button needs katsu+gateway changes, which are provided in the bento `releases/v13.1` branch


gsfk (commented):
I'll treat as a katsu issue whether this is exactly the response needed (eg why is "dataset" there?)

Otherwise looks good with just one request... is it possible to do download without "target=blank"? These are typically very small files, so this gives a really annoying screen flash on download.

gsfk (approved):

PR #301, opened 2023-09-17
[Would not be merged, Progress check] Auth lib

SanjeevLakhwani:
Created a separation b/w bento-web and auth-lib. After approval, would move this to a new package in a new repository.
(Do not comment on the commit messages)


src/lib/auth/src/performAuth.js
```diff
@@ -52,7 +49,7 @@ export const setLSNotSignedIn = () => {
     localStorage.removeItem(LS_BENTO_WAS_SIGNED_IN);
 };
 
-export const useHandleCallback = (callbackPath, authCodeCallback = undefined) => {
+export const useHandleCallback = (callbackPath, fetchUserDependentData, authCodeCallback = undefined) => {
     const dispatch = useDispatch();
```
davidlougheed, line 53: will this work in a typescript project where we'd expect an app dispatch?

src/lib/auth/src/utils.js
```diff
@@ -53,3 +53,25 @@ export const useHasResourcePermission = (resource, permission) => {
...
+    }
+};
+
+export const popLocalStorageItem = key => {
+    const val = localStorage.getItem(key);
+    localStorage.removeItem(key);
+    return val;
+};
+
+export const nop = () => {};
```
davidlougheed, line 77: missing final newline

src/lib/auth/src/performAuth.js
```diff
@@ -26,22 +23,22 @@ export const createAuthURL = async (authorizationEndpoint, scope = "openid email
...
         code_challenge: await pkceChallengeFromVerifier(verifier),
         code_challenge_method: "S256",
     }).toString();
 };
 
 const DEFAULT_REDIRECT = "/overview";
 
-export const performAuth = async (authorizationEndpoint, scope = "openid email") => {
-    window.location = await createAuthURL(authorizationEndpoint, scope);
+export const performAuth = async (authorizationEndpoint, clientId, authCallbackURL,  scope = "openid email") => {
```
davidlougheed, line 37: im wondering if this function could be wrapped inside a hook, so that it could have access to the state / config provider

```js
export const usePerformAuth = () => {
  // get authorizationEndpoint, clientId, authCallbackURL from config - maybe scope too!
  return (scope = "openid email") => {
    window.location = await createAuthURL(authorizationEndpoint, clientId, authCallbackURL, scope);
  }
}

const performAuth  = usePerformAuth();  // as a hook, this gets access to the config
// now performAuth has the signature (scope) => void;  - it still needs to be called to actually do anything
```

src/lib/auth/src/redux/configSlice.js
```diff
@@ -0,0 +1,20 @@
...
+        setConfig: (state, action) => {
+            state.CLIENT_ID = action.payload.CLIENT_ID;
+            state.AUTH_CALLBACK_URL = action.payload.AUTH_CALLBACK_URL;
+            state.OPENID_CONFIG_URL = action.payload.OPENID_CONFIG_URL;
+        }
+    },
+});
+
+export const { setConfig } = configSlice.actions;
+export default configSlice.reducer;
```
davidlougheed, line 20: missing final newline

davidlougheed (changes requested):
some suggestions

src/lib/auth/src/performAuth.js
SanjeevLakhwani: I looked it up, it cannot

davidlougheed: why can it not?

SanjeevLakhwani: it has to be within a node to access it

PR #302, opened 2023-09-19
fix(manager): if katsu cannot be contacted, show an error in the projects/datasets page

davidlougheed:
It used to just show 'no projects', which is misleading if Katsu is not running.

Now it shows the following if Katsu cannot be contacted:
Screenshot 2023-09-19 at 3 51 37 PM


gsfk (approved):
Works well! Overview still hangs forever in the same situation, but that's easily a separate issue.

PR #303, opened 2023-09-20
feat(explorer): new table/single-descriptions-based linkable experiments view

davidlougheed:
also includes
* dataset-based resource fetching to link ontologies correctly
* ontology links from biosample search results

requires the `releases/v13.1` branch of the main Bento repo, with Katsu 4.2.0!!


gsfk (commented):
Looks good and works as expected. 

One aspect of this that we've already noted is how the links tend to highlight existing flaws in the data. For my data in particular no biosample tissue links were working: they either were missing from resources (OBI:0000407 in this case) or the links were broken (NCIT:C13325, which points to a bad link: http://purl.bioontology.org/ontology/NCIT/C13325 ... is the resources information incorrect?)

PR #307, opened 2023-10-04
feat: more UI-displayable file types for drop box + experiment results

davidlougheed:
the following file types should be now viewable in drop box and in experiment results:

* CSV
* TSV
* images (JPEG, PNG, GIF, etc.)
* markdown
* audio files (MP3, M4A)
* video files (MP4)
* DOCX
* XLS/XLSX


src/components/display/FileDisplay.js
```diff
@@ -0,0 +1,275 @@
...
+const BASE_PDF_OPTIONS = {
+    cMapUrl: "cmaps/",
+    cMapPacked: true,
+    standardFontDataUrl: "standard_fonts/",
+};
+
+const LANGUAGE_HIGHLIGHTERS = {
+    "bash": "bash",
+    "json": "json",
+    "md": "markdown",  // TODO: component to either render this or show source
```
gsfk, line 56: Isn't this TODO done?

gsfk (approved):
Looks good, tried a bunch of file types with no real issues. 

The fixed-width modal makes a few files uglier than they could be, but I don't think that matters.

src/components/display/FileDisplay.js
davidlougheed: yes, ty

PR #312, opened 2023-10-12
feat(manager): add grants/groups viewer

davidlougheed:
No permissions editing yet, but lets grants/groups be visualized and interacted with, somewhat.


src/components/manager/access/PermissionsList.js
```diff
@@ -0,0 +1,55 @@
...
+            {sortedPermissions.slice(0, showAll ? sortedPermissions.length : 4).map((p) => (
+                <li key={p}>
+                    <Typography.Text code={true}>{p}</Typography.Text>
+                </li>
+            ))}
+            {sortedPermissions.length > 4 ? (
+                showAll ? (
+                    <li><a href="#" onClick={onCollapse}>- Collapse</a></li>
+                ) : (
+                    <li><a href="#" onClick={onShowAll}>+ {sortedPermissions.length - 4} more</a></li>
```
v-rocheleau, line 45: I would declare a constant like `const MAX_COLLAPSED_PERMISSIONS = 4`, since we are reusing the value.

src/components/manager/access/Subject.js
```diff
@@ -0,0 +1,76 @@
...
+        return (
+            <p style={{ margin: 0, lineHeight: "1.6em" }}>
+                <strong>{sub ? "Subject" : "Client"}:</strong>{" "}
+                <Typography.Text code={true}>{sub ?? client}</Typography.Text><br />
+                <strong>Issuer:</strong>{" "}
+                <Typography.Text code={true}>{iss}</Typography.Text><br />
+                {(sub === currentIDToken?.sub && iss === currentIDToken?.iss) ? <><em>(this is you)</em></> : null}
+            </p>
+        );
+    } else if (group) {
```
v-rocheleau, line 25: In which case would a subject have no `sub` or `client`, but would have a `group`?

v-rocheleau (changes requested):

src/components/manager/access/Subject.js
davidlougheed: there are four subject possibilities from the bento authorization service:

```
{iss, sub}
{iss, client}
{group: #}
{everyone: true}
```

in this case, the group definition itself contains a list of iss+(sub/client) or an expression which is evaluated on the token

src/components/manager/access/PermissionsList.js
davidlougheed: good point

noctillion (approved):
lgtm! Changes work as expected.

src/components/manager/access/ManagerAccessContent.js
```diff
@@ -0,0 +1,25 @@
...
+
+const ManagerAccessContent = () => {
+    const { path } = useRouteMatch();
+    return (
+        <Layout>
+            <Layout.Content style={LAYOUT_CONTENT_STYLE}>
+                <Switch>
+                    <Route path={`${path}/grants`}><AccessTabs /></Route>
+                    <Route path={`${path}/groups`}><AccessTabs /></Route>
+                    <Route path={path} render={() => <Redirect to={`${path}/grants`} />} />
```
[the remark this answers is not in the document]
davidlougheed: if a user goes to `/.../access`, this then redirects them to the first tab, `/.../access/grants`. Similarly, if in the URL a subpath that doesn't exist is specified, it'll redirect to the first tab. It's not strictly necessary but it's a way of having links elsewhere to the "access zone" from which we can choose the default sub-page and change it in the future if for example we added an `overview` tab.

v-rocheleau (approved):
lgtm!

PR #322, opened 2023-10-30
style(manager): sticky drop box control bar

davidlougheed:


davidlougheed:
@gsfk its up on qa if u dont feel like pulling it locally

gsfk (approved):
Nice! And subtle enough that I completely failed to notice it the first few times.

PR #323, opened 2023-10-31
fix(explorer): experiment result view button conditions + bad file view style

davidlougheed:


src/components/explorer/IndividualExperiments.js
```diff
@@ -52,24 +58,29 @@ const ExperimentResultActions = ({ result }) => {
...
-                    <FileDisplay uri={url} fileName={result.filename} />
-                )}
-            </Modal>
-            <Button size="small" icon="eye" onClick={onViewClick}>View</Button>{" "}
+                title={<span>View: {result.filename}</span>}
+                url={url}
+                fileName={result.filename}
+                hasTriggered={hasTriggeredViewModal}
+            />
+            <Tooltip title="View">
```
noctillion, line 81: The button is not triggering the file download as expected. Could there be a conflict between the tooltip events and the onClick handler?

noctillion (changes requested):

src/components/explorer/IndividualExperiments.js
davidlougheed: might be the case, i'll check it out!

davidlougheed: should be fixed now

noctillion (approved):
lgtm

PR #325, opened 2023-11-13
refact: factor out auth code into bento-auth-js library

SanjeevLakhwani:


src/components/App.js
```diff
@@ -108,7 +111,7 @@ const App = () => {
             const { code, verifier } = e.data ?? {};
             if (!code || !verifier) return;
             localStorage.removeItem(LS_SIGN_IN_POPUP);
-            dispatch(tokenHandoff(code, verifier));
+            dispatch(tokenHandoff({ code, verifier, CLIENT_ID, AUTH_CALLBACK_URL }));
```
davidlougheed, line 114: once again - parameters should be not in constants-case, since in tokenHandoff it isnt known if theyre truly constants

src/components/SiteHeader.js
```diff
@@ -22,32 +27,26 @@ const LinkedLogo = React.memo(() =>
...
+    </Link>
+));
 
-const CustomHeaderText = React.memo(() =>
-    <h1 style={{ color: "rgba(255, 255, 255, 0.95)", float: "left", margin: "0 24px 0 0" }}>
-        {CUSTOM_HEADER}
-    </h1>,
-);
+const CustomHeaderText = React.memo(() => (
+    <h1 style={{ color: "rgba(255, 255, 255, 0.95)", float: "left", margin: "0 24px 0 0" }}>{CUSTOM_HEADER}</h1>
```
davidlougheed, line 34: can you turn off your auto-formatter that formats files with rules we don't enforce in bento_web? it makes it hard to review this code.

src/lib/auth/redux/authSlice.js
```diff
@@ -0,0 +1,232 @@
...
+                    localStorage.setItem(LS_BENTO_WAS_SIGNED_IN, "true");
+                }
+            })
+            .addCase(refreshTokens.rejected, (state, { payload }) => {
+                const { error, error_description: errorDesc } = payload.data ?? {};
+                const tokensRefreshError = error
+                    ? `${error}: ${errorDesc}`
+                    : `An error occurred while refreshing tokens: ${action.caughtError ?? "Unknown error"}`;
+                console.error(tokensRefreshError);
+                console.error(tokensRefreshError);
```
davidlougheed, line 187: double console error

src/lib/auth/redux/openIdConfigSlice.js
```diff
@@ -0,0 +1,49 @@
...
+    initialState,
+    reducers: {},
+    extraReducers: (builder) => {
+        builder.addCase(fetchOpenIdConfiguration.pending, (state) => {
+            state.isFetching = true;
+        });
+        builder.addCase(fetchOpenIdConfiguration.fulfilled, (state, { payload }) => {
+            state.isFetching = false;
+            if (payload) {
+                console.log(payload);
```
davidlougheed, line 35: remove

davidlougheed (changes requested):
overall looking pretty close

src/components/SiteHeader.js
SanjeevLakhwani: This one is an outcome of the printWidth property, which it doesnt violate but prettier uses to reformat to best fit code

src/lib/auth/redux/authSlice.js
```diff
@@ -0,0 +1,231 @@
...
+            console.error("Missing authorization service");
+            return;
+        }
+        const key = makeResourceKey(resource);
+        const rp = thunkAPI.getState().auth.resourcePermissions[key];
+        if (rp?.isFetching || rp?.permissions) {
+            return;
+        }
+        // TODO: ask for it
+        const url = `${thunkAPI.getState().services.itemsByKind.authorization.url}/policy/permissions`;
```
davidlougheed, line 61: this needs to be done before everything is factored out

src/lib/auth/redux/authSlice.js
```diff
@@ -0,0 +1,231 @@
...
+        const response = await fetch(url, {
+            method: "POST",
+            headers: { "Content-Type": "application/json" },
+            body: JSON.stringify({ requested_resource: resource }),
```
davidlougheed, line 65: this system has changed a bit, see latest changes to bento web - we'll need to incorporate that

davidlougheed (changes requested):

src/lib/auth/redux/authSlice.js
```diff
@@ -0,0 +1,245 @@
...
+        return await response.json();
+    }
+);
+
+export const refreshTokens = createAsyncThunk(
+    "auth/REFRESH_TOKENS",
+    async ({ clientId }, { getState }) => {
+        if (!getState().auth.refreshToken) return;
+
+        const url = getState().openIdConfiguration.data?.["token_endpoint"];
```
davidlougheed, line 34: check if we have `token_endpoint` in condition, and if we don't, return false in the condition + add a console.error saying that we're missing `token_endpoint`/openID config data. then, we don't need `data?.` since we should just error if we don't have the endpoint.

src/lib/auth/redux/authSlice.js
davidlougheed: this isn't resolved, not sure why it was marked as resolved

src/lib/auth/redux/authSlice.js
```diff
@@ -0,0 +1,245 @@
...
+                state.resourcePermissions[key] = {
+                    ...state.resourcePermissions[key],
+                    isFetching: false,
+                    hasAttempted: true,
+                    permissions: payload?.result ?? [],
+                };
+            })
+            .addCase(fetchResourcePermissions.rejected, (state, { meta, payload, error }) => {
+                const key = makeResourceKey(meta.arg.resource);
+                console.error(error);
```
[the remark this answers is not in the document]
davidlougheed: (missing conditional)

davidlougheed (changes requested):

src/lib/auth/redux/authSlice.js
SanjeevLakhwani: welp, now it is

src/utils/misc.js
```diff
@@ -10,23 +10,9 @@ export const countNonNullElements = (arr) => {
 }; // counts the non-null elements in any array
 
 // Object utilities
-export const simpleDeepCopy = o => JSON.parse(JSON.stringify(o));
+export const simpleDeepCopy = (o) => JSON.parse(JSON.stringify(o));
 export const objectWithoutProps = (o, ps) => Object.fromEntries(Object.entries(o).filter(([p2]) => !ps.includes(p2)));
 export const objectWithoutProp = (o, p) => objectWithoutProps(o, [p]);
-export const arrayToObjectByProperty = (l, p) => Object.fromEntries(l.map(l => [l[p], l]));
```
davidlougheed, line 16: i use this in my ref genome PR, don't remove it please

src/store.js
```diff
@@ -0,0 +1,69 @@
+import { configureStore } from "@reduxjs/toolkit";
+import { readFromLocalStorage, writeToLocalStorage } from "./utils/localStorageUtils";
+import rootReducer from "./reducers";
+import thunkMiddleware from "redux-thunk";
+import { LS_OPENID_CONFIG_KEY } from "bento-auth-js"
```
davidlougheed, line 5: order these imports starting with libraries, then bento auth js, then local imports

davidlougheed (changes requested):
as usual, this PR also has way too many formatting changes, but i'll ignore that since we're in a rush.

src/components/App.js
```diff
@@ -2,6 +2,15 @@ import React, { Suspense, lazy, useRef, useState, useEffect, useCallback } from
...
 import { Redirect, Route, Switch, useHistory } from "react-router-dom";
 import { ChartConfigProvider } from "bento-charts";
+import {
+    fetchOpenIdConfiguration,
+    createAuthURL,
+    useHandleCallback,
+    getIsAuthenticated,
+    refreshTokens,
+    tokenHandoff,
+    LS_SIGN_IN_POPUP,
```
davidlougheed, line 12: this was moved out, but the actual callbacks related to popup handling were not - might be nice to consolidate in the future

davidlougheed (approved):

davidlougheed (changes requested):

src/components/App.js
v-rocheleau: WIP: https://github.com/bento-platform/bento_auth_js/pull/2

src/components/App.js
```diff
@@ -197,52 +173,27 @@ const App = () => {
...
-                dispatch(fetchUserDependentData(nop));
-            });
-            sessionWorker.current = sw;
-        }
-    }, [sessionWorker]);
+    useSessionWorkerTokenRefresh(
+        CLIENT_ID,
+        sessionWorker,
+        createSessionWorker,
+        fetchUserDependentData(nop),
```
davidlougheed, line 186: this is going to result in constant re-running of the effect; it should be in a useMemo or since it doesnt depend on anything inside, factored out into a constant

davidlougheed (changes requested):

src/components/App.js
```diff
@@ -76,42 +66,28 @@ const App = () => {
...
-    })();
+    const isInAuthPopup = checkIsInAuthPopup(BENTO_URL_NO_TRAILING_SLASH);
+
+    const openIdConfig = useOpenIdConfig(OPENID_CONFIG_URL);
 
     // Set up auth callback handling
-    useHandleCallback(CALLBACK_PATH, isInAuthPopup ? popupOpenerAuthCallback : undefined);
+    useHandleCallback(
+        CALLBACK_PATH,
+        fetchUserDependentData,
```
davidlougheed, line 81: having the `nop` here wasn't redundant - it was needed since fetchUserDependentData in bento web is an action factory, not an action itself

davidlougheed (changes requested):

src/components/App.js
v-rocheleau: Removed nop since it was redundant anyway.

davidlougheed: I don't think it was - the action factory takes a function callback argument, and it won't work as an action itself - it needs to be passed something to work as a callback (this could probably be set as a default argument in the action factory instead, but either way `fetchUserDependentData` won't work since it's a factory, not an action)

src/components/App.js
v-rocheleau: `fetchUserDependentData` accepts undefined `servicesCb`, in which case it will use `nop` inside the thunk. Or am I missing something?

davidlougheed: then it should be `fetchUserDependentData()` not `fetchUserDependentData`, since `fetchUserDependentData` is a factory that produces an action (and I think `useHandleCallback` expects an action, not an action factory)

src/components/App.js
v-rocheleau: Sorry for my brain fart, I missed that `fetchUserDependentData` was a thunk action creator. 
You're right, I changed it in both places and the inner thunk function is in a constant now.

src/hooks.js
```diff
@@ -30,3 +33,61 @@ export const useWorkflows = () => {
...
+
+    return {
+        fetchingPermission,
+        hasPermission,
+    };
+};
+
+/**
+ * Returns the user's permissions for a given resource
+ * @param {string} resource The resource (e.g. "everything")
```
davidlougheed, line 76: wrong jsdoc - resource should be an object like `{"everything": true}` or `{"project": "..."}`

src/hooks.js
```diff
@@ -30,3 +33,61 @@ export const useWorkflows = () => {
...
+ * The user permissions for a given resource
+ * @typedef {Object} ResourcePermissions
+ * @property {string[]} permissions The list of permissions the user has on the resource
+ * @property {boolean} isFetchingPermissions Indicates if the permissions are being fetched.
+ * @property {boolean} hasAttemptedPermissions Indicates if a permissions fetch was attempted.
+ */
+
+/**
+ * Evaluate if the user has a permission on a given resource
+ * @param {Object} resource The resource key (e.g. "everything")
```
davidlougheed, line 59: example resource in text is wrong; it's not a string

davidlougheed (changes requested):

davidlougheed (approved):
seems to work ok! `useResourcePermissions` fetches permissions many times sometimes, since a dispatch is always called when the hook is initialized, so maybe that should be fixed in `bento_auth_js`.

davidlougheed:
@v-rocheleau if you merge master into here, I'll give this a final test

davidlougheed (approved):
lgtm

PR #326, opened 2023-11-15
[15] feat: reload overview summary when a WES run completes

davidlougheed:


gsfk (approved):
I was finally able to test this one and seems to work as expected.

PR #329, opened 2023-11-22
refact(manager): rewrite run list as functional component + tweak style

davidlougheed:
should behave exactly the same as before (run list, that is)


gsfk (approved):
works as expected. 

I don't really understand the use of the hardcoded refresh time (RUN_REFRESH_TIMEOUT), is this still doing anything?

davidlougheed:
@gsfk its just how long it waits between refreshing runs. basically, if the websocket connection can't be made for some reason, this ensures we still occasionally refresh the runs while they're running + update status.

PR #330, opened 2023-11-30
refact(explorer): use context instead of redux for explorer individual info

davidlougheed:
realized this AM that this is a better way of avoiding prop-drilling in this situation, since we don't need this state globally, just in children of the explorer component

also factors out explorer individual URL building + fixes some hook usage


v-rocheleau (approved):
Lgtm! Somehow I forgot about react context, good idea to use it here.

PR #331, opened 2023-12-01
Fix dataset data-types summaries

v-rocheleau:
**Issue:** The `datasetSummaries` state was not updated properly, causing new state keys to erase old ones when it shouldn't. Katsu summaries would first be saved correctly, then the gohan summaries would erase the old state since it was not spread.

**Fix:** spread the new value with the oldstate if its an object.


davidlougheed (approved):
lgtm

PR #334, opened 2023-12-06
fix: treatment dose intervals table

v-rocheleau:
- Replace removed `renderTimeInterval` func with `TimeInterval` component
- Add a `br` prop to `TimeInterval` for optional inline/breakline
- Fix missing table key warning


davidlougheed (approved):
lgtm!

PR #337, opened 2023-12-12
feat(manager): genome list view

davidlougheed:


src/components/explorer/IndividualTracks.js
```diff
@@ -165,7 +165,12 @@ const IndividualTracks = ({ individual }) => {
         referenceGenomes.items.forEach((g) => {
             availableGenomes[g.id] = { id: g.id, fastaURL: g.fasta, indexURL: g.fai };
         });
-        (igvGenomes.data ?? []).forEach((g) => availableGenomes[g.id] = g);
+        (igvGenomes.data ?? []).forEach((g) => {
+            availableGenomes[g.id] = g;
+            // Manual aliasing for well-known human genome aliases:
+            if (g.id === "hg19") availableGenomes["GRCh37"] = {...g, id: "GRCh37"};
+            else if (g.id === "hg38") availableGenomes["GRCh38"] = {...g, id: "GRCh38"};
```
v-rocheleau, line 172: Would it be worth it to have the reference service store these aliases instead? As we support non-human genomes the list of aliases could grow. 
Alternatively, we could have some sort of genomes_aliases.json config file instead of hard coding here.

v-rocheleau (approved):
Lgtm, suggestion for aliases

src/components/explorer/IndividualTracks.js
davidlougheed: the ref service can store aliases; this is because for now we want to take advantage of igv.js's gene tracks, but they only store GRCh37/38 as hg19/38. In the future, this can be removed since we will ingest GFF3 files into the reference service. I'll add a comment explaining this

gsfk: Can report that for mice in particular using "mm10" as an assembly gets me the "RefSeq Genes" track from igv, but using "GRCm38" doesn't, even though they're the same assembly.

davidlougheed: ah i will add that then and see if igv.js has a built-in list of aliases

gsfk (commented):
Looks good with just a couple comments:
- prod vcf metadata will have to be updated to include assembly id field (arguably this should have been done already anyhow). 
- the "View track" toggle hangs while track is loading

davidlougheed:
hanging should be fixed now; also I added mouse genome aliases for the igv.js annotations

gsfk:
all looks good now.

gsfk (approved):

PR #338, opened 2023-12-19
fix biosample procedure props and rendering

v-rocheleau:
- Prop Type for `biosample.procedure` was mistakenly mandatory, now optional.
- BiosampleProcedure now returns an em dash if the `procedure` prop or its `code` is falsy


gsfk (approved):
Works fine; `code` is required in procedure, so the second falsy check has no effect (I can confirm that procedures without codes fail at ingest).

PR #339, opened 2024-01-04
feat: provide link to biosample from genomic interpretations if applicable

v-rocheleau:
- If a `GenomicInterpretation` has a `subject_or_biosample_id` that is related to a biosample, we provide a link to the biosample
- Otherwise, if `subject_or_biosample_id` is related to a subject, we display the ID in plain text, since we are already on the individual's explorer page


davidlougheed (approved):
lgtm - i think if we have computed extra properties we should give them a prefix though (like `__` or something unique that we reserve - and then we should prevent user-extra properties from having keys starting with that prefix) - otherwise it's unclear which extra props are computed vs which are not.

davidlougheed (approved):
lgtm

PR #340, opened 2024-01-05
fix datasetResources reducer

v-rocheleau:
**Issue:** the `datasetResources` reducer did not properly handle array values, causing them to be spread in an object instead of an array in the state.

### DIFF
--- a/src/components/explorer/utils.js
+++ b/src/components/explorer/utils.js
@@ -100,7 +100,7 @@ export const useDatasetResources = (datasetIDOrDatasetIDs) => {
             const r = Object.values(
                 Object.fromEntries(
                     datasetIDs
-                        .flatMap(d => datasetResources[d]?.data ?? [])
+                        .flatMap(d => Object.values(datasetResources[d]?.data ?? []))
                         .map(r => [r.id, r]),
                 ),
             );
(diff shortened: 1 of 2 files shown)


src/components/explorer/utils.js
```diff
@@ -100,7 +100,7 @@ export const useDatasetResources = (datasetIDOrDatasetIDs) => {
             const r = Object.values(
                 Object.fromEntries(
                     datasetIDs
-                        .flatMap(d => datasetResources[d]?.data ?? [])
+                        .flatMap(d => Object.values(datasetResources[d]?.data ?? []))
```
davidlougheed, line 103: I don't really understand this - it worked before, and dataset resources should be a list of resources to my knowledge.

davidlougheed (changes requested):

src/components/explorer/utils.js
davidlougheed: it can also be an object, but then it doesn't match with the metadata resources field defined in phenopackets - if it truly is an object, the default value should be `{}` not `[]`

v-rocheleau: The issue was not the hook after all, but the reducer. 
I removed this change and corrected the reducer so that it handles arrays properly.

davidlougheed (approved):
ah thats what it was...... lgtm

PR #343, opened 2024-01-11
fix(manager): out-of-order WES event handling

davidlougheed:


src/modules/wes/actions.js
```diff
@@ -21,10 +21,18 @@ export const fetchRuns = networkAction(() => (dispatch, getState) => ({
...
 
-export const receiveRunDetails = (runID, data) => ({
+/**
+ * Manually dispatch a run details receive event, equivalent to what is fired by the network action.
+ * @param {string} runID
+ * @param {Object} data
+ * @param {number | undefined} ts
+ * @return {{data, runID, type: string, ts: number | undefined}}
+ */
+export const receiveRunDetails = (runID, data, ts = undefined) => ({
```
v-rocheleau, line 31: Minor: I find `ts` to be ambiguous for an function argument name, it would be more clear if we name it `timestamp` I think.

src/modules/wes/reducers.js
```diff
@@ -65,9 +65,11 @@ const streamError = (state = INITIAL_RUNS_STATE, action, stream) => {
 const makeRunSkeleton = (run, request) => ({
     ...run,
     state: "QUEUED",  // Default initial state
+    ts: null,  // Will get replaced with a UTC timestamp when we receive updates or events
```
v-rocheleau, line 68: Minor: rename to `timestamp` to be more explicit

v-rocheleau (approved):
Lgtm! Minor naming suggestion

v-rocheleau (approved):

PR #346, opened 2024-01-23
fix(overview): clickable Other + needless extra code for Other

davidlougheed:
this ended up becoming a bit of a refactor because the overview code is quite repetitive and strangely constructed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -121,19 +121,6 @@
       "resolved": "https://registry.npmjs.org/@ant-design/icons/-/icons-2.1.1.tgz",
       "integrity": "sha512-jCH+k2Vjlno4YWl6g535nHR09PwCEmTBKAG6VqF+rhkrSPRLfgpU2maagwbZPLjaHuU5Jd1DFQ2KJpQuI6uG8w=="
     },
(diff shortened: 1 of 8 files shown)


gsfk:
Note that our prod data on occasion has a category with the literal name "Other"

davidlougheed:
ah, good to know, thanks.

gsfk (approved):
Works as expected, I did my best to break it. 

Only issue I see is that there is no longer any difference between the grouped "Other" category and pie segments literally named "Other", although that seems to be an existing bento charts issue, rather than an issue with this pr. "Other" occurs in several places in katsu's experiment schema, for example (I couldn't find one in prod, I may have been wrong about that). 

So a good future ticket would be to go back to some of the ideas about distinguishing the grouped "Other" from the remaining segments in a bar chart (the tooltip could show the number of categories grouped together, or their names, or the segment could be distinguished in some obvious way). And then perhaps we could replace the feature that allowed you to search categories literally named "Other", although again that might be a bento charts matter. 

I'd love to see a cBioPortal-style chart component that is user-configurable as pie, bar, or table, which might eliminate the need for this sort of thing.

PR #349, opened 2024-01-25
fix: missed state selector updates from auth-js refact

davidlougheed:
@SanjeevLakhwani if you refactor the redux state, please remember to search for all uses of the selected state and correctly update them.


v-rocheleau (approved):
Oups, lgtm

PR #352, opened 2024-01-31
patch study_types pie chart

gsfk:
Experiments "study type" pie chart was showing experiment type (experiment type was displayed twice)


davidlougheed (approved):
lgtm, oops

PR #354, opened 2024-01-31
fix: misc errors that are showing up in console

davidlougheed:


gsfk (approved):
looks so empty now with a blank console....

PR #364, opened 2024-02-07
"negated" -> "excluded" for phenotypic features

gsfk:
Change phenopackets v1 term "negated" to v2 term "excluded" for phenotypic features.

### DIFF
--- a/src/components/explorer/IndividualPhenotypicFeatures.js
+++ b/src/components/explorer/IndividualPhenotypicFeatures.js
@@ -10,7 +10,7 @@ const PHENOTYPIC_FEATURES_COLUMNS = [
     {
         title: "Feature",
         key: "feature",
-        render: ({ header, type, negated }) => ({
+        render: ({ header, type, excluded }) => ({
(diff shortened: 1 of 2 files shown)


src/components/explorer/IndividualPhenotypicFeatures.js
```diff
@@ -20,7 +20,7 @@ const PHENOTYPIC_FEATURES_COLUMNS = [
                 </h4>
             ) : <>
                 <OntologyTerm term={type} />{" "}
-                {negated ? (
+                {excluded ? (
```
davidlougheed, line 23: can you also add this property (`excluded`) to `propTypes.js` (which is barely used, but just for completeness?)

gsfk: this is a good point, I already ran across a [propTypes issue](https://206.12.92.46/issues/2069) in testing this pr

davidlougheed (approved):
lgtm

PR #367, opened 2024-02-26
fix(propTypes): biosample sampled_tissue no longer required

gsfk:
Remove `isRequired` from biosample sampled_tissue prop types, no longer required in Phenopackets 2.0

Partial fix for Redmine [#2069](https://206.12.92.46/issues/2069).


davidlougheed (approved):
lgtm

PR #372, opened 2024-03-05
refact(manager): project JSON Schema form to ant4 + schema input

davidlougheed:
instead of a bunch of state passing for the json schema input, this creates a custom Ant input component for the JSON schema data, which makes it a bit cleaner (in addition to the port to Ant 4).

### DIFF
--- a/src/components/manager/projects/ProjectJsonSchemaForm.js
+++ b/src/components/manager/projects/ProjectJsonSchemaForm.js
@@ -38,7 +38,7 @@ const JsonSchemaInput = ({ value, onChange }) => {
                     // Validate against draft-07 meta schema
                     onChange(json);
                 } else {
-                    message.error("Selected file is an invalid JSON schema definition.").catch(console.error);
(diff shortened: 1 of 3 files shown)


src/components/manager/projects/ProjectJsonSchemaForm.js
```diff
@@ -56,101 +54,83 @@ const ProjectJsonSchemaForm = ({ style, schemaTypes, initialValues, setFileConte
...
+                        <span>Check to make the <ExtraPropertiesCode tooltip/> field required</span>
+                    }>
+                        <span>Required</span>
+                    </Tooltip>
+                }
+                name="required"
+                initialValue={initialValues.required}
+                valuePropName="checked"
+            >
+                <Checkbox />
```
v-rocheleau, line 115: Minor: since the `required` field is the only non-mandatory form item, we could place it at the end of the form for consistency, instead of in the middle.

v-rocheleau (approved):
lgtm, minor suggestion

PR #373, opened 2024-03-05
refact(explorer): convert discovery search form to ant 4

davidlougheed:


src/components/discovery/DiscoverySearchCondition.js
```diff
@@ -34,217 +48,164 @@ export const getSchemaTypeTransformer = type => {
...
+        if (onChange) onChange(newState);
+    }, [fieldState, onChange]);
 
-        this.handleChange = this.handleChange.bind(this);
-    }
+    const { field, fieldSchema, operation, negated, searchValue } = fieldState;
 
-    handleField = (value, key = "field", fieldSchemaKey = "fieldSchema") => {
-        if (this.state[key] === value.selected) return;
+    const handleField = useCallback((value) => {
```
v-rocheleau, line 84: I would use a different name for the value callback argument, since "value" is one of the component's props too

davidlougheed:
need to fix: clicking from charts doesn't work for Julian

src/components/discovery/DiscoverySearchCondition.js
v-rocheleau, line 1: Degradation in the condition validation.
Previously the explorer would validate that a search value exists for a condition.
Now the user can submit a search request with empty condition values, then fail silently.

v-rocheleau (changes requested):

src/components/discovery/DiscoverySearchForm.js
```diff
@@ -34,180 +33,180 @@ const VARIANT_OPTIONAL_FIELDS = [
...
+    const isEnum = fieldSchema.hasOwnProperty("enum");
+
+    // noinspection JSCheckFunctionSignatures
+    if (
+        !VARIANT_OPTIONAL_FIELDS.includes(field) &&
+        (transformedSearchValue === null ||
+            (!isEnum && !transformedSearchValue) ||
+            (isEnum && !fieldSchema.enum.includes(transformedSearchValue)))
+    ) {
+        return Promise.reject(`This field is must have a value: ${field}`);
```
v-rocheleau, line 57: Phrasing

v-rocheleau: Validation: we may want to prevent users from submitting white-space search values

v-rocheleau (changes requested):
Looks good and behaves as expected, besides two minor comments

noctillion (approved):
Looks good to me. Changes work as expected.

v-rocheleau (approved):
lgtm!

PR #374, opened 2024-03-07
[WIP] refact: rewrite routes to react router 5.1 style

davidlougheed:


src/components/RequireAuth.js
```diff
@@ -0,0 +1,62 @@
...
+    }
+
+    return isAuthenticated ? children : (
+        <Layout.Content style={styles.layout}>
+            <Empty
+                image={<SignInIcon />}
+                imageStyle={styles.emptyImage}
+                description="You must sign into this node to access this page."
+            >
+                {isAuthenticated ? (
```
v-rocheleau, line 46: Won't this always be `false`, given that we are in the falsy part of the parent `isAuthenticated` ternary?

v-rocheleau (commented):
Lgtm and behaves as expected, one question

src/components/RequireAuth.js
davidlougheed: true, this was just straight ported from the old OwnerRoute but it's redundant for sure.

PR #383, opened 2024-03-18
refact: replace react-json-view with react-18-compat lib

davidlougheed:
also moves react18-json-view usage into fewer locations

PR #384, opened 2024-03-19
refact: React 18, React Router 6, bento_charts 2.6, -victory

davidlougheed:


gsfk (commented):
Lots of stuff to check, here's a few things I found so far: 

- When using click-to-search from the overview page, sometimes (1/20-30 times?) the search condition does not appear on the search page, although the search works normally otherwise. I couldn't reproduce this bug in current master. 


- clicking the services list gives this error: 
> ```
> ServiceList.js:116 Selector unknown returned a different result when called with the same parameters. This can lead to unnecessary rerenders.
> Selectors that return a new reference (such as an object or an array) should be memoized
> ```

- The alignment between pies and bar charts is a little wonky (they seem to be aligned at the top) although I think current master is the same.

src/components/charts/PieChart.js
noctillion, line 1: The pie size is smaller than usual. Not sure if this is intentional.

davidlougheed: should be a bit better now

src/components/manager/runs/RunDetailContent.js
```diff
@@ -15,38 +14,29 @@ const styles = {
...
 
     // TODO: 404
     const run = runsByID[id] || null;
     const loading = (run?.details ?? null) === null;
 
-    const onChangeTab = useCallback(
-        (key) => history.push(`/admin/data/manager/runs/${run?.run_id}/${key}`),
-        [history, run]);
-    const onBack = useCallback(() => history.push("/admin/data/manager/runs"), [history]);
+    const onChangeTab = useCallback((key) => navigate(`../${key}`), [navigate, run]);
```
v-rocheleau, line 27: We can remove `run` from dependency array I think

v-rocheleau (changes requested):
Looks pretty good, small comments. Didn't notice any degradations

src/components/explorer/ExplorerIndividualContent.js
```diff
@@ -74,75 +71,77 @@ const ExplorerIndividualContent = () => {
...
-    const biosamplesUrl = `${individualUrl}/biosamples`;
-    const experimentsUrl = `${individualUrl}/experiments`;
-    const diseasesUrl = `${individualUrl}/diseases`;
-    const ontologiesUrl = `${individualUrl}/ontologies`;
-    const tracksUrl = `${individualUrl}/tracks`;
-    const phenopacketsUrl = `${individualUrl}/phenopackets`;
-    const interpretationsUrl = `${individualUrl}/interpretations`;
-    const medicalActionsUrl = `${individualUrl}/medical-actions`;
-    const measurementsUrl = `${individualUrl}/measurements`;
+    const individualUrl = `/data/explorer/individuals/${individualID}`;
```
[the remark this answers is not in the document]
davidlougheed: I'm not sure how this would end up looking...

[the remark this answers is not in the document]
v-rocheleau: Yeah me neither, this is fine I think

gsfk (commented):
More stuff, many possibly due to other recent prs, sorry ( first one is probably from #383)

- Json view behaviour has changed a bit, some missing stuff is not handled gracefully, eg when experiments don't have extra_properties: 
Screenshot from 2024-03-27 16-09-46

- Download button seems broken for me, I get the same error from the download button in Data Manager > DRS Objects and the one in Explorer > (some individual) > Experiments

```
DownloadButton.js:55 Uncaught TypeError: Cannot convert undefined or null to object
    at Function.entries (<anonymous>)
    at DownloadButton.js:55:16
    at handleClick (button.js:185:1)
    at HTMLUnknownElement.callCallback (react-dom.development.js:188:1)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:237:1)
    at invokeGuardedCallback (react-dom.development.js:292:1)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:306:1)
    at executeDispatch (react-dom.development.js:389:1)
    at executeDispatchesInOrder (react-dom.development.js:411:1)
    at executeDispatchesAndRelease (react-dom.development.js:3278:1)
    at executeDispatchesAndReleaseTopLevel (react-dom.development.js:3287:1)
    at forEachAccumulated (react-dom.development.js:3259:1)
    at runEventsInBatch (react-dom.development.js:3304:1)
    at runExtractedPluginEventsInBatch (react-dom.development.js:3514:1)
    at handleTopLevel (react-dom.development.js:3558:1)
```
- from Drop Box, clicking multiple vcfs and selecting the vcf ingestion workflow does not work correctly, and throws this error: `Warning: value should be an array when TreeSelect is checkable or multiple.` However, I don't think I've ever tried this before, so may already have been broken. 
- hard reloads briefly flash the "You must sign into this node to access this page" component (possibly a "don't care" issue)
- I'm not convinced I should be seeing all these search options, at least not without better names: 
wuttt
- The individual "Phenopackets JSON" tab throws the same selector error that the services page was showing before: 
`Selector unknown returned a different result when called with the same parameters.`

davidlougheed:
1) will be fixed by https://github.com/bento-platform/bento_web/pull/387
2) fixed
3) fixed
4) needs changes to bento auth JS, will be addressed though
5) fixed
6) fixed
7) fixed
8) fixed
9) oh no :(

davidlougheed:
9 should be fixed now

gsfk:
Okay! Everything above looks resolved. In verifying, I found one more issue and lots of warnings:

- in Drop Box, everything has a "folder" icon whether it's a directory or a document (I might have missed this one before)

warnings:
- in Drop Box, when viewing an xlsx file: `Warning: [antd: Table] index parameter of rowKey function is deprecated. There is no guarantee that it will work as expected. `
- when clicking (some individual) > Phenotypic Features: `Warning: columns.render return cell props is deprecated with perf issue, please use onCell instead.`
- in search, in some situations typing in the input box produces this: `Warning: There may be circular references`

gsfk:
Only just now trying variants search, which gives me this: `Warning: Return type is option instead of Option instance. Please read value directly instead of reading from props.  LocusSearch.js:88 `

gsfk:
Ah, much more serious issue: in queries with multiple data types, it looks like the variant part of the query is ignored (never sent in the request)

gsfk:
Variant queries seem to launch now, although search conditions are forgotten when clicking away and returning to the page.

davidlougheed:
@gsfk as far as i can tell this is existing behaviour unfortunately

gsfk:
huh

davidlougheed:
at least in ICHANGE it is if i fill out the form and tab to manager and back. maybe in some cases there's a regression though?

src/components/discovery/VariantSearchHeader.js
```diff
@@ -8,22 +8,39 @@ import LocusSearch from "./LocusSearch";
...
 import { notAlleleCharactersRegex } from "@/utils/misc";
 
-const VariantSearchHeader = ({dataType, addVariantSearchValues}) => {
-    const [refFormReceivedValidKeystroke, setRefFormReceivedValidKeystroke ] = useState(true);
-    const [altFormReceivedValidKeystroke , setAltFormReceivedValidKeystroke ] = useState(true);
+
+const isValidLocus = (locus) => {
+    return locus.chrom !== null && locus.start !== null && locus.end !== null;
+};
+const validateAlleleText = (text) => {
```
v-rocheleau, line 15: Rename to something like `formatAlleleText` since it doesn't do validation

src/components/display/CsvDisplay.js
```diff
@@ -54,7 +53,7 @@ const CsvDisplay = ({ contents, loading }) => {
             loading={loading || isParsing}
             columns={columns}
             dataSource={parsedData}
-            rowKey={rowKey}
+            rowKey="__key__"
```
v-rocheleau, line 56: Store this string in a constant and use it at L28 as well

src/components/display/XlsxDisplay.js
```diff
@@ -64,7 +63,7 @@ const XlsxDisplay = ({ contents }) => {
                 scroll={TABLE_SCROLL}
                 columns={sheetColumns}
                 dataSource={sheetJSON}
-                rowKey={rowKey}
+                rowKey="__key__"
```
v-rocheleau, line 66: idem

v-rocheleau (changes requested):
Small comments, otherwise lgtm.
For me the explorer's form is preserved after navigating to a different tab and back.

gsfk:
> at least in ICHANGE it is if i fill out the form and tab to manager and back. maybe in some cases there's a regression though?

Yes, pretty sure this worked sometime in the past.

davidlougheed:
I agree the Workflows tab is redundant now. Perhaps I'll remove it in a new PR

davidlougheed:
the circular reference thing doesn't seem to break much, it's annoying but I don't know what's causing it so maybe I'll just leave it.

gsfk (approved):
working now for me

PR #385, opened 2024-03-20
fix(manager): dataset form (fix it not working)

davidlougheed:
refact(manager): rewrite dataset form as functional
fix(manager): deprecation warning for dataset form
style(manager): fix spacing between project datasets

### DIFF
--- a/src/components/datasets/DatasetForm.js
+++ b/src/components/datasets/DatasetForm.js
@@ -6,15 +6,15 @@ const { Item } = Form;
 
 import DataUseInput from "../DataUseInput";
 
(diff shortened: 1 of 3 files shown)


gsfk (commented):
I can now add a new dataset, but editing the dataset details is still broken. "Save" does nothing, and throws this console error:

```
DatasetFormModal.js:51 TypeError: Cannot read properties of undefined (reading 'props')
    at DatasetFormModal.js:43:34
```

v-rocheleau (approved):
Lgtm, able to add and update now

gsfk (approved):
works for me too

PR #390, opened 2024-04-04
feat(explorer): show biosample ID with experiment data

davidlougheed:


gsfk (commented):
Works well. I was only slightly confused by having biosample id as the first column, since it looks like the biosample table over again. Maybe experiment type first?

gsfk (approved):

PR #392, opened 2024-04-05
refact: handle new Katsu overview format

davidlougheed:


v-rocheleau (approved):
lgtm, but we should wait for the Katsu PR to be merged first

gsfk (approved):

PR #394, opened 2024-04-10
refact: remove deprecated Ant API usages

davidlougheed:


gsfk (approved):
Looks fine. 

Can't fully test `ProjectJsonSchemaForm.js` since it seems like the katsu feature responsible for it seems to be broken.

PR #395, opened 2024-04-15
site footer: link bento logo to bento page

gsfk:
Give the "Powered by Bento" image a clickable link to https://bento-platform.github.io/. 

Pops out in a new page (`target="_blank"`) even though the other footer links don't do this. I'm happy to hear contrary suggestions. 

`rel="noreferrer"` now mostly voodoo, only necessary for a few very old browers which probably can't run Bento.


davidlougheed (approved):
lgtm. if the bento website URL is ever used elsewhere we should probably make that a constant, but it looks good for now.

PR #397, opened 2024-04-17
feat: permissions awareness (part 1)

davidlougheed:
some basic UI permissions awareness in place. lots more fine-grain work + dataset permissions + UI for grants/groups needed.


src/components/RequireAuth.js
```diff
@@ -21,36 +21,28 @@ const SignInIcon = React.memo(() => (
...
-                imageStyle={styles.emptyImage}
-                description="You must sign into this node to access this page."
-            >
-                <Button type="primary" loading={openIdConfigNotLoaded} onClick={performAuth}>
-                    Sign In
-                </Button>
-            </Empty>
-        </Layout.Content>
+    return (
+        <AutoAuthenticate>
```
v-rocheleau, line 31: neat

src/components/AutoAuthenticate.js
v-rocheleau, line 1: Make new files .ts

src/modules/wes/hooks.js
v-rocheleau, line 1: Make new files .ts

src/modules/services/hooks.js
v-rocheleau, line 1: Make new files .ts

src/modules/notifications/hooks.js
v-rocheleau, line 1: Make new files .ts

v-rocheleau (changes requested):
first pass

nginx.conf
```diff
@@ -8,6 +8,11 @@ http {
     include /etc/nginx/mime.types;
     default_type application/octet-stream;
 
+    # Performance tweaks for static content serving
+    #  - See https://docs.nginx.com/nginx/admin-guide/web-server/serving-static-content/
+    sendfile on;
+    tcp_nopush on;
```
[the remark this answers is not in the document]
davidlougheed: no, just read up on some best practices for static file serving.

src/modules/metadata/hooks.js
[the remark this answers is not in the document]
davidlougheed: won't bother for anything needing dispatch/state typing

src/components/SiteHeader.js
```diff
@@ -73,24 +85,49 @@ const SiteHeader = () => {
...
+                text: "Reference Genomes",
+                key: "genomes",
             },
+            ...(managerPermissions.canManageAnything ? [
+                {
+                    key: "data-manager",
+                    url: "/data/manager",
+                    icon: <FolderOpenOutlined />,
+                    text: "Data Manager",
+                    disabled: !isAuthenticated || !managerPermissions.canManageAnything,
```
gsfk, line 103: not sure `disabled` is doing anything here:
- managerPermissions.canManageAnything is always true if we made it to this code, so right side of disjunction is always false.
- I'm guessing left side is always false for the same reasons

src/components/SiteHeader.js
```diff
@@ -73,24 +85,49 @@ const SiteHeader = () => {
...
+                },
+                // For now, only show the services page to users who can manage something, since it's not useful for
+                // end users.
+                {
+                    key: "services",
+                    url: "/services",
+                    icon: <DashboardOutlined />,
+                    text: "Services",
+                    disabled: !isAuthenticated,
```
gsfk, line 112: same reasoning here as above, will never be `disabled` since this menu item only appears conditionally

gsfk (commented):
Working well will only a few issues, some may be permissions issues rather than specific to this pr. 


Regarding the site header menu items, the display is a bit inconsistent about whether you're allowed to use that menu item or not:

- always appears even when inaccessible (Overview)
- greyed out when inacessible? (Explorer?)
- shows a loading gif, then the item appears or doesn't appear at all (Data Manager / Services) .. a few inline comments below on this one. 



On testing, I got a few behaviours that are possibly intentional, and others than probably aren't:

When completely logged out: 

- I can modify the settings (the control for pie chart threshold) ... looks like this has been possible for a while. 
- I can see but not alter reference genomes
- I can't access Overview or Explorer, but "Explorer" in the menu is greyed out, while "Overview" isn't


I created a new user in Keycloak and logged in, but without managing permissions at all in authz (not sure which behaviour this is supposed to produce)

- "Explorer" is greyed out in the top menu, but I was still able to use the Explorer page, as well as click-to-search from the front page.

src/components/SiteHeader.js
```diff
@@ -156,11 +195,20 @@ const SiteHeader = () => {
...
-        [isAuthenticated, authHasAttempted, idTokenContents, performSignOut, unreadNotifications],
+        [
+            isAuthenticated,
+            authHasAttempted,
+            idTokenContents,
+            performSignOut,
+            performAuth,
+            canQueryData,
+            managerPermissions,
+            unreadNotifications,
```
v-rocheleau, line 210: Missing: `openIdConfigFetching`, `isHandingOffCodeForToken`, `isFetchingPermissions`, `canViewNotifications`, `hasAttemptedQueryPermissions`, `hasAttemptedManagerPermissions`.

We should probably add the eslint rules for react hooks and dependency array, this way the linter can do these checks for us.
"react-hooks/rules-of-hooks": "error", // For checking rules of hooks
"react-hooks/exhaustive-deps": "warn" // For checking hook dependencies

src/components/explorer/SearchSummaryModal.js
```diff
@@ -52,7 +52,7 @@ const renderCharts = (chartsData) => {
 const SearchSummaryModal = ({ searchResults, ...props }) => {
     const [data, setData] = useState(null);
 
-    const katsuUrl = useSelector((state) => state.services.itemsByArtifact.metadata.url);
+    const katsuUrl = useService("metadata")?.url;
```
v-rocheleau, line 55: Not related to this PR. The `katsuUrl` const is not in the useEffect's dependencies array, we probably want to check that `katsuUrl` is truthy before doing the fetch as well.

src/components/manager/ManagerAnalysisContent.js
```diff
@@ -2,15 +2,30 @@ import React from "react";
...
+    const { permissions, hasAttemptedPermissions } = useResourcePermissionsWrapper(RESOURCE_EVERYTHING);
+
+    // TODO: each workflow should have definitions for permissions scopes, so we can instead check if we can run at
+    //  least one workflow.
+
+    if (hasAttemptedPermissions && !permissions.includes(analyzeData)) {
+        return (
+            <ForbiddenContent message="You do not have permission to view the analysis wizard." />
+        );
+    }
```
v-rocheleau, line 27: Can be simplified with `useHasResourcePermissionWrapper(RESOURCE_EVERYTHING, analyzeData)`.

src/components/manager/ManagerExportContent.js
```diff
@@ -2,15 +2,30 @@ import React from "react";
...
+    const { permissions, hasAttemptedPermissions } = useResourcePermissionsWrapper(RESOURCE_EVERYTHING);
+
+    // TODO: each workflow should have definitions for permissions scopes, so we can instead check if we can run at
+    //  least one workflow.
+
+    if (hasAttemptedPermissions && !permissions.includes(exportData)) {
+        return (
+            <ForbiddenContent message="You do not have permission to view the export wizard." />
+        );
+    }
```
v-rocheleau, line 27: Can be simplified with `useHasResourcePermissionWrapper(RESOURCE_EVERYTHING, analyzeData)`.

src/components/manager/ManagerIngestionContent.js
```diff
@@ -2,15 +2,33 @@ import React from "react";
...
+    //  least one workflow.
+
+    if (
+        hasAttemptedPermissions &&
+        !(permissions.includes(ingestData) || permissions.includes(ingestReferenceMaterial))
+    ) {
+        return (
+            <ForbiddenContent message="You do not have permission to view the ingestion wizard." />
+        );
+    }
```
v-rocheleau, line 30: Suggestion: Here I feel a hook like `useHasResourcePermissions(resource: Resource, permissions: string[], checkAll=true)` would be useful. If `checkAll` is true, check that all the permissions are present, if it is false, check that at least one permission is present.

v-rocheleau (changes requested):
Hooks suggestions

src/components/SiteHeader.js
davidlougheed: good idea - i'll make a separate PR after this for that I think

src/components/explorer/SearchSummaryModal.js
davidlougheed: same as above, good point

src/components/manager/ManagerAnalysisContent.js
davidlougheed: doesn't include a `hasAttempted` attr, so it doesn't work for this case.

src/components/manager/ManagerExportContent.js
davidlougheed: same here

src/components/manager/ManagerAnalysisContent.js
davidlougheed: (should do in bento auth js to support this)

v-rocheleau: Ah true, if we add `hasAttempted` we could even wrap this in an authz component RequirePermissions, similar to RequireAuth and AutoAuthenticate.

src/components/manager/ManagerIngestionContent.js
davidlougheed: I think I'll do that in a future PR as well, if that works

src/utils/actions.js
```diff
@@ -63,18 +63,25 @@ const _networkAction =
...
 
-            const { types, params, url, baseUrl, req, err, onSuccess, onError, paginated } = fnResult;
+            const { types, check, params, url, baseUrl, req, err, onSuccess, onError, paginated } = fnResult;
+
+            // if we're currently auto-authenticating, don't start any network requests; otherwise, they have a good
+            // chance of getting interrupted when the auth redirect happens.
+            if (getState().auth.isAutoAuthenticating) return;
+
+            // if a check function is specified and evaluates to false on the state, don't fire the action.
+            if (check && !check(getState())) return;
```
v-rocheleau, line 73: great idea

src/modules/authz/hooks.js
```diff
@@ -0,0 +1,115 @@
...
+                canIngest ||
+                canAnalyzeData ||
+                canExportData ||
+                canViewRuns ||
+                canViewPermissions
+            ),
+        },
+        isFetching: isFetchingPermissions || isFetchingManageProjectsDatasetsPermissions,
+        hasAttempted: hasAttemptedPermissions && hasAttemptedManageProjectsDatasetsPermissions,
+    };
```
v-rocheleau, line 114: Maybe memoize this? This way we can skip all the `includes` operations if the permissions don't change.

src/modules/manager/actions.js
```diff
@@ -25,21 +27,22 @@ export const beginProjectEditing = basicAction(PROJECT_EDITING.BEGIN);
 export const endProjectEditing = basicAction(PROJECT_EDITING.END);
 
 
-const fetchDropBoxTree = networkAction(() => (dispatch, getState) => ({
+export const fetchDropBoxTree = networkAction(() => (dispatch, getState) => ({
     types: FETCH_DROP_BOX_TREE,
+    check: (state) => state.services.dropBoxService
+        && !state.dropBox.isFetching
+        && (!state.dropBox.tree.length || state.dropBox.isInvalidated),
```
v-rocheleau, line 34: What is this condition for?

src/modules/manager/actions.js
```diff
@@ -25,21 +27,22 @@ export const beginProjectEditing = basicAction(PROJECT_EDITING.BEGIN);
...
-    try {
-        return await dispatch(fetchDropBoxTree());
-    } catch (e) {
-        // Reset loading
-        console.error(e);
-        return await dispatch({type: FETCH_DROP_BOX_TREE.FINISH});
-    }
+export const invalidateDropBoxTree = basicAction(INVALIDATE_DROP_BOX_TREE);
+
+export const fetchDropBoxTreeIfNeededOrFail = () => async (dispatch, getState) => {
```
v-rocheleau, line 41: Is the "OrFail" suffix still true here?

v-rocheleau (changes requested):

src/modules/manager/actions.js
davidlougheed: it's another pattern of forcing a drop box (or other state) reload - instead of calling a non-checked fetch action version, we mark the state as invalidated and then whenever `fetchIfNeeded` or whatever is called, it will re-perform the fetch since the state is marked as invalid.

src/modules/manager/actions.js
davidlougheed: no, fixed

src/modules/manager/actions.js
davidlougheed: is this fine?

gsfk:
Still have a question left from last time: is it expected to get query data permissions when creating a new user in Keycloak (and  doing nothing else)?

davidlougheed:
no, users shouldn't auto-get `query:data` unless there's something (an iss/client grant perhaps) giving it to them

explorer back end still works due to katsu authz not being implemented yet I think, but i'll double check the front end.

gsfk:
With the same no-rights user I can also hand-type the /services url and get a working services page, and was able to call katsu `/individuals`...

davidlougheed:
both are expected, services don't need permissions and katsu only checks for the no-permissions user right now because no authz yet.

davidlougheed:
(thus why this is only still a part 1...)

gsfk:
ah, okay, just seemed odd that services is shown as unavailable, but is still there

davidlougheed:
it's essentially because it's not that interesting for non-managing users, but we can always make it available.

gsfk:
I have all permissions and I'm reloading the page: what's the second spinner for? 

Peek 2024-05-08 10-21

gsfk (approved):
No new comments except something we've talked about already: pie charts probably shouldn't be clickable at all if the user doesn't have query data permissions (search doesn't work for them now, but they're still encouraged to click, which takes them to a "forbidden" notice).

src/components/manager/projects/Project.js
```diff
@@ -49,10 +49,13 @@ const Project = ({
...
-    const { permissions, isFetchingPermissions } = useResourcePermissionsWrapper(resource);
+    const {
+        permissions: projectPermissions,
+        isFetchingPermissions: isFetchingProjectPermissions,
+    } = useResourcePermissionsWrapper(resource);
 
-    const canEditProject = permissions.includes(editProject);
-    const canCreateDataset = permissions.includes(createDataset);
+    const canEditProject = projectPermissions.includes(editProject);
+    const canCreateDataset = projectPermissions.includes(createDataset);
```
v-rocheleau, line 58: Minor: wrap in a useMemo

src/modules/authz/hooks.js
```diff
@@ -81,35 +81,49 @@ export const useManagerPermissions = () => {
...
-        },
-        isFetching: isFetchingPermissions || isFetchingManageProjectsDatasetsPermissions,
-        hasAttempted: hasAttemptedPermissions && hasAttemptedManageProjectsDatasetsPermissions,
-    };
+    return useMemo(() => {
+        const canViewDropBox = permissions.includes(viewDropBox);
+        const canIngest = permissions.includes(ingestData) || permissions.includes(ingestReferenceMaterial);
+        const canAnalyzeData = permissions.includes(analyzeData);
+        const canExportData = permissions.includes(exportData);
+        const canQueryData = permissions.includes(queryData);
```
v-rocheleau, line 89: Should `canQueryData` be a part of the `canManageAnything` permission? I get that it is not a "manager" permission per-se, but it is used in the manager to enable/disable the DRS tab, and `canManageAnything` exposes the data manager.

v-rocheleau (changes requested):
Looking good, only 2 questions before approving.

src/modules/authz/hooks.js
davidlougheed: the problem with it being part of `canManageAnything` is that then users which can just query data (i.e., most users) can then see the manager tab... maybe in the future we can have an explicit permission for seeing the manager tab that's separate, but as long as it's derived from permissions it makes sense to leave it out, I think

v-rocheleau: Yeah I agree with you. That or we can have a permission for DRS querying and remove canQueryData from the manager permissions.

v-rocheleau (approved):
LGTM! Though we should wait for part 2 to be merged in `feat/manager/auth-aware` before we merge in master. Starting to review part 2 now!

davidlougheed:
@v-rocheleau it may make it easier to work on other web branches in parallel if we merge this now and merge part 2 separately?

PR #400, opened 2024-04-24
feat: permissions awareness pt. 2 + access manager UI

davidlougheed:


gsfk (commented):
All the permissions-aware stuff looks good. A few general comments on the Access manager UI:

- some of this stuff is a little obscure, so needs documentation, for example I have no idea what "Expression (Bento Query Format)" is 
- I wasn't able to create a group, the "Create Group" modal gives "method not allowed" on clicking "Create" 
`bentov2-authz             | INFO:     172.26.0.2:51062 - "PUT /groups/ HTTP/1.1" 405 Method Not Allowed`
- for creating Grants, the "Project + Optional Dataset + Optional Data Type" option only lets you pick one of each, when presumably you might want to allow access to more than one project, more than one dataset, or some subset of the datatypes. 
- If you pick a datatype, you can't submit the form, since all permissions are greyed out, but the form requires at least one permission.
- shouldn't it be possible to edit existing grants? (I want to add / remove permissions for a particular user)

gsfk:
For working better now for me

gsfk (approved):

src/components/manager/access/ExpiryInput.tsx
v-rocheleau, line 1: After saving a grant with a UTC-4 (Montreal) expiry, the authz service stores the datetime it in UTC-0.
When viewing the created grant in web, it shows as UTC-0 instead of UTC-4. So there is a 4 hour offset that is confusing.
If we don't convert datetimes to the browser's timezone, we could at least show the expiry TZ to avoid confusion here.

src/components/manager/access/GrantForm.tsx
```diff
@@ -0,0 +1,514 @@
...
+        if (subject) onChange(subject);
+    }
+};
+
+const SubjectInput = ({ value, onChange }: SubjectInputProps) => {
+    const groups: StoredGroup[] = useGroups().data;
+
+    const homeIssuer = useOpenIdConfig()?.issuer ?? "";
+
+    const [subjectType, setSubjectType] = useState<"everyone" | "iss-sub" | "iss-client" | "group">("everyone");
```
v-rocheleau, line 74: use type instead of literals: `useState<SubjectType>("everyone")`.

src/components/manager/access/GrantForm.tsx
```diff
@@ -0,0 +1,514 @@
...
+
+
+const SUBJECT_EVERYONE: GrantSubject = { everyone: true };
+
+type SubjectInputProps = {
+    value?: GrantSubject;
+    onChange?: (v: GrantSubject) => void;
+};
+
+type SubjectType = "everyone" | "iss-sub" | "iss-client" | "group";
```
v-rocheleau, line 35: Literals are repeated a bunch, putting them in consts would help us avoid typos and ease refactors in the future.

src/components/manager/access/GrantForm.tsx
```diff
@@ -0,0 +1,514 @@
...
+};
+
+const ResourceInput = ({ value, onChange }: ResourceInputProps) => {
+    // TODO: consolidate when useProjects() is typed
+    const projects: Project[] = useProjects().items;
+    const projectsByID: Record<string, Project> = useProjects().itemsByID;
+    const datasetsByID: Record<string, Dataset> = useProjects().datasetsByID;
+    const dataTypes: BentoServiceDataType[] = useDataTypes().items;
+
+    const [resourceType, setResourceType] = useState<"everything" | "project-plus">("everything");
```
v-rocheleau, line 202: Idem for string literals, types and consts would be better I think.

src/components/manager/access/GrantForm.tsx
```diff
@@ -0,0 +1,514 @@
...
+    const [selectedDataType, setSelectedDataType] = useState<string>("");
+
+    useEffect(() => {
+        if (!value) return;
+        if ("everything" in value) {
+            setResourceType("everything");
+        } else {
+            // TODO: how to handle missing projects?
+            setSelectedProject(projectsByID[value.project]);
+            // TODO: how to handle missing datasets?
```
v-rocheleau, line 214: Not sure I get the TODOs. If "everything" is not in `value`, then project should be defined according to `Resource` type.

v-rocheleau (changes requested):
Quick look focusing on the grant form.

src/components/manager/access/GrantForm.tsx
davidlougheed: this means "what if the project has been deleted"

davidlougheed: (right now, it doesn't matter since we cannot edit grants)

src/components/manager/access/ExpiryInput.tsx
davidlougheed: the `Z` trailing char means UTC, but i agree that it's not user-friendly.

davidlougheed: added a proper locale rendering component

src/components/manager/access/Subject.tsx
```diff
@@ -0,0 +1,93 @@
...
+                <Link to={`/data/manager/access/groups#group-${group}`}>
+                    {groupDef
+                        ? (<>{groupDef.name} (ID: {group})</>)
+                        : (<>ID: {group}</>)}
+                </Link>
+            </>
+        );
+    } else if ("everyone" in subject) {
+        return (
+            <Popover content="Everyone, even anonymous users."><span style={labelStyle}>Everyone</span></Popover>
```
v-rocheleau, line 62: What use cases would we want to cover with grants assigned to `everyone`?
Assuming `everyone` has a grant to delete projects, would it be possible for an anonymous user to delete a project?
I'm worried about anonymous users having dangerous permissions.

v-rocheleau (changes requested):

src/components/manager/access/GrantsTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+        },
+        {
+            title: "Resource",
+            dataIndex: "resource",
+            render: (resource) => <Resource resource={resource} />,
+        },
+        {
+            title: "Expiry",
+            dataIndex: "expiry",
+            render: (expiry) => <span>{expiry ?? "—"}</span>,
```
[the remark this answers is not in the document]
davidlougheed: good idea

src/components/manager/access/GrantForm.tsx
```diff
@@ -0,0 +1,514 @@
...
+    const newValueSet = new Set([...newChecked]);
+    const checkedSet = new Set([...checked]);
+
+    const difference1 = new Set(checked.filter((c) => !newValueSet.has(c)));
+    const difference2 = new Set(newChecked.filter((c) => !checkedSet.has(c)));
+
+    return !!difference1.size || !!difference2.size;
+};
+
+const permissionCompatibleWithResource = (p: PermissionDefinition, r: GrantResource) => {
```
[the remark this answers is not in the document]
davidlougheed: yes, that's why that's there

src/components/manager/access/GrantsTabContent.js
```diff
@@ -1,103 +1,124 @@
...
+            title: "Actions",
+            key: "actions",
+            // TODO: hook up edit
+            render: (grant) => {
+                const pObj = grantResourcePermissionsObjects[makeResourceKey(grant.resource)];
+                const pLoading = pObj.isFetching;
+                const canEdit = pObj.permissions.includes(editPermissions);
+                return (
+                    <>
+                        {/*TODO: no edit grant right now; originally designed to be immutable but this should change*/}
```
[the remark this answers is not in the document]
davidlougheed: no, since it requires changes to authz too

src/components/manager/access/Subject.tsx
davidlougheed: maybe i should put a big warning? and maybe we can forbid some permissions from being everyone-enabled... but for example, RENATA wants all data to be publicly discoverable, and eventually we should have all discovery endpoints use the bool/counts permissions

src/components/manager/access/GrantsTabContent.js
```diff
@@ -1,103 +1,124 @@
...
+
+const GrantCreationModal = ({ open, closeModal }) => {
+    const dispatch = useAppDispatch();
+    const [form] = Form.useForm();
+
+    const onOk = useCallback(() => {
+        form.validateFields().then(async (values) => {
+            console.debug("received grant values for creation:", values);
+            await dispatch(createGrant(values));
+            closeModal();
```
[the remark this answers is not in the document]
davidlougheed: already done

src/components/manager/access/GroupsTabContent.tsx
```diff
@@ -0,0 +1,309 @@
...
+                <>
+                    <Button size="small" icon={<EditOutlined />} disabled={editModalOpen} onClick={() => {
+                        setSelectedGroup(group);
+                        setEditModalOpen(true);
+                    }}>Edit</Button>{" "}
+                    <Button
+                        size="small"
+                        danger={true}
+                        icon={<DeleteOutlined />}
+                        onClick={() => {
```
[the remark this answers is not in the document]
davidlougheed: doesn't make sense here since it's inside a render function in table columns, i think

src/components/manager/access/GrantsTable.tsx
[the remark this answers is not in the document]
davidlougheed: done

src/components/manager/access/Subject.tsx
davidlougheed: I've added a warning, but maybe it should also not be the default choice...

src/components/manager/access/GroupsTabContent.tsx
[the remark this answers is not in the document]
v-rocheleau: Oh right nvm

v-rocheleau (changes requested):
Last comments I think, looks good!

src/modules/authz/hooks.js
```diff
@@ -73,6 +146,53 @@ export const useCanQueryAtLeastOneProjectOrDataset = () =>
...
+export const useAuthzManagementPermissions = () => {
+    const { data: grants, isFetching: isFetchingGrants } = useGrants();
+
+    // Get existing project/dataset resource permissions
+    const projectDatasetPermissions = useProjectDatasetPermissions();
+
+    // Build set of deduplicated grant resources
+    const grantResources = useMemo(
+        () => [...new Set([
+            RESOURCE_EVERYTHING_KEY,
```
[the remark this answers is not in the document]
davidlougheed: added a comment explaining.

src/modules/authz/types.ts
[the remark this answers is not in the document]
davidlougheed: added a TODO

v-rocheleau (approved):
Lgtm, nice work on the forms!

PR #401, opened 2024-05-03
Update IndividualVariants.js

noctillion:
fix pass symbol to setIgvPosition


davidlougheed (approved):
matches spec.

gsfk (changes requested):
Bit of a puzzle here for me. which I didn't notice when I wrote this ticket. 

The relevant field of phenopackets is [GeneDescriptor](https://phenopacket-schema.readthedocs.io/en/latest/gene.html#rstgene) (aka "gene_context") but the code in web here seems to be for [variant expression](https://phenopacket-schema.readthedocs.io/en/latest/variant.html#rstexpression), which does not have a `gene_context` field (doesn't have `value_id` or `symbol` either)

So the relevant component to change is actually [GeneDesciptor.js](https://github.com/bento-platform/bento_web/blob/master/src/components/explorer/IndividualGenes.js#L31-L48). And we should consider removing the dead code, unless as is always possible, I completely misunderstood.

Also a smaller quibble that the clickable link should actually show the symbol and not the value_id.

davidlougheed:
hmm yeah its true that this is actually odd despite being functional... @v-rocheleau what is the context here for the duplicate gene context passing?

gsfk:
Looking at the code it seems like gene_context is pulled from variationDescriptor.gene_context

src/components/explorer/IndividualVariants.js
```diff
@@ -26,28 +20,13 @@ const variantExpressionPropType = PropTypes.shape({
...
 
-const VariantExpressionDetails = ({variantExpression, geneContext}) => {
-    const dispatch = useDispatch();
-    const { individualID } = useContext(ExplorerIndividualContext);
-    const tracksUrl = useMemo(() => {
-        if (individualID) {
-            return `${explorerIndividualUrl(individualID)}/tracks`;
-        }
-    }, [individualID]);
+const VariantExpressionDetails = ({variantExpression}) => {
```
v-rocheleau, line 23: Can be wrapped in a `memo`

v-rocheleau: Also modify the rendering code to remove the `gene_context` prop (line 64) and conditional render (line 61)

v-rocheleau (changes requested):
There was no good reason for the duplicate `gene_context` when I implemented it, other than my poor understanding of igv.

v-rocheleau (approved):
Lgtm

PR #402, opened 2024-05-06
feat(manager): add column indicating if DRS perms resource is valid

davidlougheed:


gsfk (approved):
Does everything I expect it to do, deletes orphaned files correctly.

PR #404, opened 2024-05-13
feat: scoped discovery config

v-rocheleau:


src/modules/services/hooks.js
```diff
@@ -1,6 +1,9 @@
 import { useEffect, useMemo } from "react";
 import { useDispatch, useSelector } from "react-redux";
 import { fetchBentoServices, fetchServices } from "@/modules/services/actions";
```
davidlougheed, line 3: this should be changed to a local import, since it's same-level

src/modules/metadata/actions.js
```diff
@@ -84,7 +84,7 @@ const _fetchExtraPropertiesSchemaTypes = networkAction(() => (dispatch, getState
     error: "Error fetching extra properties schema types",
 }));
 
-export const fetchExtraPropertiesSchemaTypes = () => (dispatch, getState) => {
+export const fetchExtraPropertiesSchemaTypes = (dispatch, getState) => {
```
davidlougheed, line 87: true that this doesn't need to be a double-function anymore, but it does make it a bit inconsistent with other actions in bento...  hmm

src/modules/discovery/actions.js
```diff
@@ -18,3 +17,15 @@ const performGohanGeneSearch = networkAction((searchUrl) => () => ({
...
     err: "error performing Gohan gene search",
 }));
+
+const _fetchDiscoverySchema = networkAction(() => (dispatch, getState) => ({
+    types: FETCH_DISCOVERY_SCHEMA,
+    url: `${getState().bentoServices.itemsByKind.metadata.url}/api/discovery_schema`,
+    err: "Error fetching discovery JSON schema",
+}));
+
+export const fetchDiscoverySchema = (dispatch, getState) => {
```
davidlougheed, line 27: this module was eliminated in the perms pt. 2 PR, so it'll lead to conflicts - we can reintroduce it, or move this to the metadata module.

src/hooks.ts
```diff
@@ -1,9 +1,19 @@
...
+import {
+    RESOURCE_EVERYTHING,
+    useAuthorizationHeader,
+    useHasResourcePermission,
+    useResourcePermissions,
+} from "bento-auth-js";
 import type { Resource } from "bento-auth-js";
 
 import { useService } from "@/modules/services/hooks";
+import { RootState } from "./store";
```
davidlougheed, line 12: `import type`

src/hooks.ts
```diff
@@ -1,9 +1,19 @@
...
+import { useCallback, useEffect, useState } from "react";
```
davidlougheed, line 13: weird import ordering - move this to top

src/hooks.ts
```diff
@@ -1,9 +1,19 @@
...
+import Ajv from "ajv";
```
davidlougheed, line 14: move above local imports

src/components/manager/projects/ProjectCreationModal.js
```diff
@@ -1,42 +1,41 @@
-import React, { useCallback, useRef } from "react";
+import React, { useCallback } from "react";
 import { useSelector, useDispatch } from "react-redux";
 import { useNavigate } from "react-router-dom";
 
-import { Button, Modal } from "antd";
+import { Button, Modal, Form } from "antd";
```
davidlougheed, line 5: alpha order for imports

src/components/manager/projects/Project.js
```diff
@@ -231,7 +243,7 @@ const Project = ({
                             </Empty>
                         )
 
-                    }
+                }
```
davidlougheed, line 246: odd formatting change

src/components/manager/projects/Project.js
```diff
@@ -75,25 +78,34 @@ const Project = ({
...
+
+
 
     const handleSave = useCallback(() => {
-        const form = editingForm.current;
-        if (!form) return;
-        form.validateFields().then((values) => {
+        editingForm.validateFields().then((values) => {
             // Don't save datasets since it's a related set.
+            console.log(values);
```
davidlougheed, line 91: rm or give better debug logging context text + log level

src/components/manager/projects/Project.js
```diff
@@ -75,25 +78,34 @@ const Project = ({
...
             onSave({
                 identifier: projectState.identifier,
                 title: values.title || projectState.title,
                 description: values.description || projectState.description,
                 data_use: values.data_use || projectState.data_use,
+                discovery: newDiscoveryContent || values.discovery,
             });
+            editingForm.resetFields();
```
davidlougheed, line 99: might want to await onsave, or reset form on open instead of on save to prevent the flicker of the form back to the reset values

src/components/manager/projects/Project.js
```diff
@@ -1,23 +1,23 @@
...
 import ProjectJsonSchema from "./ProjectJsonSchema";
-import { useHasResourcePermissionWrapper, useResourcePermissionsWrapper } from "@/hooks";
+import { useDropBoxFileContent, useHasResourcePermissionWrapper, useResourcePermissionsWrapper } from "@/hooks";
 import {
     createDataset,
     deleteProject,
     editProject,
     makeProjectResource,
     RESOURCE_EVERYTHING,
 } from "bento-auth-js";
```
davidlougheed, line 19: maybe move auth-js imports above local imports?

src/components/manager/projects/Project.js
```diff
@@ -1,23 +1,23 @@
-import React, { useCallback, useEffect, useMemo, useRef, useState } from "react";
+import React, { useCallback, useEffect, useMemo, useState } from "react";
 import PropTypes from "prop-types";
 
-import { Button, Col, Empty, Row, Space, Tabs, Typography } from "antd";
+import { Button, Col, Empty, Row, Space, Tabs, Typography, Form } from "antd";
```
davidlougheed, line 4: alpha order

src/components/manager/DropBoxTreeSelect.js
```diff
@@ -1,8 +1,13 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useEffect } from "react";
 import PropTypes from "prop-types";
-import { TreeSelect } from "antd";
+import { TreeSelect, Form } from "antd";
```
davidlougheed, line 3: alpha order

src/components/manager/DropBoxTreeSelect.js
```diff
@@ -1,8 +1,13 @@
-import React, { useMemo } from "react";
+import React, { useMemo, useEffect } from "react";
```
davidlougheed, line 1: alpha order

src/components/datasets/DatasetFormModal.js
```diff
@@ -23,22 +23,24 @@ const DatasetFormModal = ({ project, mode, initialValue, onCancel, onOk, open })
         isSavingDataset: projectDatasetsSaving,
     } = useProjects();
 
-    const formRef = useRef(null);
+    // const formRef = useRef(null);
```
davidlougheed, line 26: rm

src/components/datasets/DatasetFormModal.js
```diff
@@ -1,8 +1,8 @@
-import React, { useCallback, useRef } from "react";
+import React, { useCallback } from "react";
 import { useDispatch } from "react-redux";
 import PropTypes from "prop-types";
 
-import { Button, Modal } from "antd";
+import { Button, Modal, Form } from "antd";
```
davidlougheed, line 5: alpha order

src/components/datasets/DatasetForm.js
```diff
@@ -18,38 +20,53 @@ const validateJson = (rule, value) => {
...
+const DatasetForm = ({ initialValue, form, updateMode}) => {
+    const discoveryValidator = useDiscoveryValidator();
     return (
-        <Form ref={formRef} layout="vertical">
+        <Form form={form} layout="vertical" initialValues={initialValue}>
             <Item
                 label="Title"
                 name="title"
-                initialValue={initialValue?.title || ""}
+                // initialValue={initialValue?.title || ""}
```
davidlougheed, line 30: does this break editing, or is it handled by the form obj?

src/components/datasets/DatasetForm.js
```diff
@@ -18,38 +20,53 @@ const validateJson = (rule, value) => {
...
                 rules={[{ required: true }, { min: 3 }]}
             >
                 <Input placeholder="My Dataset" size="large" />
             </Item>
             <Item
                 label="Description"
                 name="description"
-                initialValue={initialValue?.description || ""}
+                // initialValue={initialValue?.description || ""}
```
davidlougheed, line 38: same q here

src/components/App.js
```diff
@@ -60,6 +60,9 @@ const App = () => {
     const dispatch = useDispatch();
     const navigate = useNavigate();
 
+    useMetadataDependentData();
```
davidlougheed, line 63: is the metadata dependent data actually used here? it seems to be not related to the contents of this component

src/modules/services/hooks.js
```diff
@@ -58,3 +61,23 @@ export const useWorkflows = () => {
         return { workflowsByType, workflowsLoading };
     }, [serviceWorkflows, workflowsLoading]);
 };
+
+
+export const useMetadataDependentData = () => {
```
davidlougheed, line 66: i think it would be better to have two hooks which actually return the redux state of the corresponding data; `useExtraPropertiesSchemaTypes` and `useDiscoverySchema`, which themselves can call the fetch if katsu is ready. this creates a cleaner path from data-dependency to the fetch action.

src/modules/services/hooks.js
```diff
@@ -58,3 +61,23 @@ export const useWorkflows = () => {
...
+    const dispatch = useDispatch();
+    const metadata = useBentoService("metadata");
+    useEffect(() => {
+        if (!metadata?.url) return;
+        dispatch(fetchExtraPropertiesSchemaTypes);
+        dispatch(fetchDiscoverySchema);
+    }, [metadata, dispatch]);
+};
+
+export const useGohanDependentData = () => {
```
davidlougheed, line 76: same as with katsu - would be better to connect this hook to where gohan variants overview data is used.

src/modules/services/hooks.js
```diff
@@ -58,3 +61,23 @@ export const useWorkflows = () => {
...
 };
+
+
+export const useMetadataDependentData = () => {
+    const dispatch = useDispatch();
+    const metadata = useBentoService("metadata");
+    useEffect(() => {
+        if (!metadata?.url) return;
+        dispatch(fetchExtraPropertiesSchemaTypes);
+        dispatch(fetchDiscoverySchema);
```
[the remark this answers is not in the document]
v-rocheleau: no longer double functions indeed

src/modules/discovery/actions.js
v-rocheleau: moved into metadata

src/components/datasets/DatasetForm.js
v-rocheleau: Removed the commented code. Initial values are handled by the form object using the Item's name, except for the two DropBoxJsonSelect components, which has its own initial value mechanism.

src/components/datasets/DatasetForm.js
v-rocheleau: Removed the commented code. Initial values are handled by the form object using the Item's name, except for the two DropBoxJsonSelect components, which has its own initial value mechanism.

src/components/manager/projects/RoutedProject.js
```diff
@@ -55,9 +64,8 @@ const RoutedProject = () => {
   }, []);
 
   const handleProjectSave = useCallback(
-    (newProject) => {
-      // TODO: Form validation for project
-      dispatch(saveProjectIfPossible(newProject));
+    async (newProject) => {
+      await dispatch(saveProjectIfPossible(newProject));
```
davidlougheed, line 68: this could just be `(newProject) => dispatch(saveProjectIfPossible(newProject))` to directly return the promise

src/components/manager/projects/Project.js
```diff
@@ -65,28 +71,31 @@ const Project = ({
...
-        console.error(err);
+  const handleSave = useCallback(async () => {
+    try {
+      const values = await editingForm.validateFields();
+      await onSave({
+        identifier: projectState.identifier,
+        title: values.title || projectState.title,
+        description: values.description || projectState.description,
+        data_use: values.data_use || projectState.data_use,
+        discovery: newDiscovery || values.discovery,
```
davidlougheed, line 87: this line is a bit confusing to me - why does it fall back to values, when everything else falls back to project state? what's the purpose of newDiscovery? (maybe in my PR I forgot to check this part... since it feels like it would have made more sense with the original logic)

src/components/manager/dropBox/DropBoxJsonSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+  initialValue?: JSONType;
+  value?: JSONType;
+  onChange?: (x: JSONType) => void;
+};
+
+/** A form input component for DropBox JSON file selection. */
+const DropBoxJsonSelect = ({ initialValue, onChange }: DropBoxJsonSelectProps) => {
+  const editing = initialValue !== undefined;
+
+  const [radioValue, setRadioValue] = useState<"existing" | "new">(editing ? "existing" : "new");
```
davidlougheed, line 41: can add the `none` value as previously discussed - be careful of some ternary logic below which relies on this being a two-value type

davidlougheed (changes requested):

src/components/manager/dropBox/DropBoxJsonSelect.tsx
```diff
@@ -28,44 +28,59 @@ const InnerDropBoxJsonSelect = ({
   );
 };
 
+const EXISTING = "existing";
+const NEW = "new";
+const NONE = "none";
+type DROP_BOX_SELECT_TYPE = typeof EXISTING | typeof NEW | typeof NONE;
```
davidlougheed, line 34: types are generally done not in all uppercase, so i'd expect this to be more like `DropBoxSelectType`

davidlougheed: maybe a string enum would work? https://www.typescriptlang.org/docs/handbook/enums.html#string-enums

src/components/manager/projects/Project.js
v-rocheleau: Line was bad. Thankfully your changes make DropBoxJsonSelect behave like a proper form item, so the hacky hooks for `newDiscovery` are no longer needed here since the form handles everything now.
I tried to do it this way initially but I missed the required props, so your PR is very welcome haha.

src/components/manager/dropBox/DropBoxJsonSelect.tsx
v-rocheleau: good idea

src/components/datasets/DatasetForm.js
```diff
@@ -1,56 +1,43 @@
...
-  try {
-    JSON.parse(value);
-    return Promise.resolve();
-  } catch (e) {
-    return Promise.reject("Please enter valid JSON");
+const DatasetForm = ({ initialValue, form }) => {
+  const discoveryValidator = useDiscoveryValidator();
+  const datsValidator = useDatsValidator();
+  if (initialValue && !initialValue?.data_use) {
+    initialValue["data_use"] = simpleDeepCopy(INITIAL_DATA_USE_VALUE);
```
davidlougheed, line 19: we can't mutate object props, since it violates the react data flow (and might have weird effects up the tree) - i think the true used initial value should be memoized, especially with a deep copy present.

src/components/manager/dropBox/DropBoxJsonSelect.tsx
```diff
@@ -0,0 +1,94 @@
...
+      multiple={false}
+      nodeEnabled={dropBoxTreeNodeEnabledJson}
+      allowClear={true}
+      value={selectedFile}
+      onChange={(v) => setSelectedFile(v)}
+    />
+  );
+};
+
+enum DropBoxSelectType {
```
davidlougheed, line 31: nice

src/components/manager/dropBox/DropBoxTreeSelect.tsx
```diff
@@ -35,7 +43,13 @@ const generateFileTree = (directory, valid, folderMode, basePrefix) =>
       };
     });
 
-const DropBoxTreeSelect = forwardRef(({ folderMode = false, nodeEnabled, basePrefix, ...props }, ref) => {
+export type DropBoxTreeSelectProps = Omit<TreeSelectProps, "showSearch" | "treeDefaultExpandAll" | "treeData"> & {
+  folderMode?: boolean;
+  nodeEnabled?: DropBoxEntryValidFunction; // TODO: more precise typing when common functions are typed
```
davidlougheed, line 48: i realized i left an outdated comment here... should be removed (oops)

davidlougheed (changes requested):

src/components/datasets/DatasetForm.js
v-rocheleau: done

davidlougheed (changes requested):
one minor thing; otherwise looks good.

PR #405, opened 2024-05-23
feat: integrate reference service 0.2.0 features

davidlougheed:


gsfk (commented):
Really nice, I like the tags in the autocomplete options... good enough that you might be able to remove `TODOs: style options` at the top of LocusSearch.js. The tag disappears when you select an option, not sure if that's deliberate. It might be worth keeping.

Just one thing is broken though. Since so many of our instances have only one assembly, the variants search form is supposed to autofill the "Assembly ID" field in the form if there's only one assembly available, rather than forcing you to go through a select menu with only one item. It seems like the form is not re-rendering.

davidlougheed:
tag disappearing I don't have much of a choice on because of how the autocomplete element works / the logic that extracts the coordinates from the text itself... auto-selecting the assembly ID should be fixed now.

gsfk (approved):
Working now. 

Weirdly it's possible to do a custom render of the selected option in antd `Select` but not antd `Autocomplete`, despite how similar they are. This isn't the first time this component has let me down.

src/components/discovery/LocusSearch.js
```diff
@@ -1,63 +1,55 @@
...
+const LocusSearch = ({ assemblyId, addVariantSearchValues, handleLocusChange, setLocusValidity }) => {
+    const referenceGenomes = useReferenceGenomes();
     const [autoCompleteOptions, setAutoCompleteOptions] = useState([]);
-    const geneSearchResults = useSelector((state) => state.discovery.geneNameSearchResponse);
     const [inputValue, setInputValue] = useState("");
-    const dispatch = useDispatch();
 
-    const showAutoCompleteOptions = assemblyId === "GRCh37" || assemblyId === "GRCh38";
+    const showAutoCompleteOptions = !!referenceGenomes.itemsByID[assemblyId]?.gff3_gz
+        && inputValue.length && (inputValue.includes(" ") || !inputValue.includes(":"));
```
v-rocheleau, line 27: minor: useMemo

src/components/explorer/IndividualTracks.js
```diff
@@ -161,9 +162,14 @@ const IndividualTracks = ({ individual }) => {
 
     const dispatch = useDispatch();
 
+    const referenceService = useService("reference");
     const igvGenomes = useIgvGenomes();  // Built-in igv.js genomes (with annotations)
     const referenceGenomes = useReferenceGenomes();  // Reference service genomes
 
+    const igvGenomesByID = useMemo(
```
v-rocheleau, line 169: Would it be worth it to have this in the redux store?

src/modules/reference/hooks.js
```diff
@@ -12,3 +13,49 @@ export const useReferenceGenomes = () => {
...
+        const params = new URLSearchParams({ name: nameQuery, name_fzy: "true", limit: "10" });
+        const searchUrl = `${referenceService.url}/genomes/${referenceGenomeID}/features?${params.toString()}`;
+
+        setError(null);
+
+        (async () => {
+            setIsFetching(true);
+
+            try {
+                const res = await fetch(searchUrl, { headers: { Accept: "application/json" } });
```
v-rocheleau, line 42: Are we sure the reference service will always be public? If not, we should pass the auth headers as well.

src/components/discovery/LocusSearch.js
```diff
@@ -1,63 +1,55 @@
...
-    const handleChange = (value) => {
+        addVariantSearchValues({ chrom, start, end });
+    }, [setLocusValidity, addVariantSearchValues]);
 
-        setInputValue(value);
-
-        if (!value.includes(" ") && value.includes(":")) {
-            handlePositionNotation(value);
+    useEffect(() => {
+        if (!inputValue.includes(" ") && inputValue.includes(":")) {
```
v-rocheleau, line 36: minor: could factor out `inputValue.includes(" ")`, `inputValue.includes(":")` into constants for reuse here and to compute `showAutoCompleteOptions`

v-rocheleau (changes requested):
looks good, small comments

v-rocheleau (approved):
lgtm!

PR #406, opened 2024-05-27
feat(explorer): search results experiment type filtering

noctillion:
This PR add filter by experiment types in experiment table in explorer section. Allow the export of filtered data to csv

### DIFF
--- a/src/components/explorer/ExplorerSearchResultsTable.js
+++ b/src/components/explorer/ExplorerSearchResultsTable.js
@@ -109,7 +109,7 @@ const ExplorerSearchResultsTable = ({
                 onSelect: () => handleSetSelectedRows([]),
             },
         ],
-    }), [selectedRows, data, handleSetSelectedRows]);
+    }), [selectedRows, filteredData, handleSetSelectedRows]);
 
     const sortedInfo = useMemo(
         () => ({
--- a/src/components/explorer/ExplorerSearchResultsTable.js
+++ b/src/components/explorer/ExplorerSearchResultsTable.js
@@ -29,22 +29,22 @@ const ExplorerSearchResultsTable = ({
 }) => {
     const { dataset } = useParams();
     const [currentPage, setCurrentPage] = useState(initialCurrentPage || 1);
-    const [activeFilter, setActiveFilter] = useState([]);
+    const [activeFilters, setActiveFilters] = useState([]);
 
     const [summaryModalVisible, setSummaryModalVisible] = useState(false);
     const [tracksModalVisible, setTracksModalVisible] = useState(false);
 
     const filteredData = useMemo(() => {
         return data.filter((item) => {
-            return Object.keys(activeFilter).every((filterKey) => {
-                const filterValues = activeFilter[filterKey] || [];
+            return Object.keys(activeFilters).every((filterKey) => {
+                const filterValues = activeFilters[filterKey] || [];
                 if (filterValues.length === 0) {
                     return true;
                 }
                 return filterValues.includes(item[filterKey]);
             });
         });
-    }, [data, activeFilter]);
+    }, [data, activeFilters]);
 
     const numResults = useMemo(() => filteredData.length, [filteredData]);
 
@@ -76,10 +76,10 @@ const ExplorerSearchResultsTable = ({
         }
     }, [dispatch, dataset, activeTab]);
 
-    const onPageChange = useCallback((pageObj, filters, sorter) => {
-        const newPage = filters !== activeFilter ? 1 : pageObj.current;
+    const onTableChange = useCallback((pageObj, filters, sorter) => {
+        const newPage = filters !== activeFilters ? 1 : pageObj.current;
         setCurrentPage(newPage);
-        setActiveFilter(filters);
+        setActiveFilters(filters);
         dispatch(setTableSortOrder(dataset, sorter.field, sorter.order, activeTab, newPage));
     }, [dispatch, dataset, activeTab]);
 
--- a/src/components/explorer/ExplorerSearchResultsTable.js
+++ b/src/components/explorer/ExplorerSearchResultsTable.js
@@ -29,10 +29,24 @@ const ExplorerSearchResultsTable = ({
 }) => {
     const { dataset } = useParams();
     const [currentPage, setCurrentPage] = useState(initialCurrentPage || 1);
-    const [numResults] = useState(data.length);
+    const [activeFilter, setActiveFilter] = useState([]);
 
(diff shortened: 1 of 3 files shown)


src/components/explorer/searchResultsTables/ExperimentsTable.js
```diff
@@ -73,14 +73,27 @@ const ExperimentsTable = ({ data, datasetID }) => {
         (state) => state.explorer.tableSortOrderByDatasetID[datasetID]?.["experiments"],
     );
 
+    const experimentTypeFilters = useDynamicFilterOptions(data, "experimentType");
+
+    const filterColumns = SEARCH_RESULT_COLUMNS_EXP.map((col) => {
+        if (col.dataIndex === "experimentType") {
```
davidlougheed, line 79: this is a confusing way to alter `SEARCH_RESULT_COLUMNS_EXP`. It'd be better to move the columns definition into a new memoized variable which has dependencies on `experimentTypeFilters`.

src/components/explorer/hooks/explorerHooks.js
```diff
@@ -29,3 +29,13 @@ export const useSortedColumns = (data, tableSortOrder, columnsDefinition) => {
 
     return { sortedData, columnsWithSortOrder };
 };
+
+export const useDynamicFilterOptions = (data, key) => {
```
davidlougheed, line 33: mention in hook name that this is for tables, like `useDynamicTableFilterOptions`

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -159,6 +175,7 @@ const ExplorerSearchResultsTable = ({
                     pagination={{
                         pageSize: PAGE_SIZE,
                         defaultCurrent: currentPage,
+                        showSizeChanger: false,
```
davidlougheed, line 178: why is this changed?

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -83,8 +99,8 @@ const ExplorerSearchResultsTable = ({
                 key: "all-data",
                 text: "Select All Data",
                 onSelect: () => {
-                    const allRowKeys = data.map((item) => item.key);
-                    handleSetSelectedRows(allRowKeys);
+                    const filteredKeys = filteredData.map((item) => item.key);
```
davidlougheed, line 102: memo dependency should be on filteredData now, not data

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -63,8 +77,10 @@ const ExplorerSearchResultsTable = ({
     }, [dispatch, dataset, activeTab]);
 
     const onPageChange = useCallback((pageObj, filters, sorter) => {
```
davidlougheed, line 79: is this called on filter change too?  in which case maybe we should call this `onTableChange` instead

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -29,11 +29,25 @@ const ExplorerSearchResultsTable = ({
 }) => {
     const { dataset } = useParams();
     const [currentPage, setCurrentPage] = useState(initialCurrentPage || 1);
-    const [numResults] = useState(data.length);
+    const [activeFilter, setActiveFilter] = useState([]);
```
davidlougheed, line 32: can this be multiple things? should it be `activeFilters` instead?

davidlougheed (changes requested):

src/components/explorer/ExplorerSearchResultsTable.js
noctillion: yes, the onPageChange function is called when filters change, in addition to handling pagination changes

src/components/explorer/ExplorerSearchResultsTable.js
noctillion: good catch! Thanks

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -29,11 +29,25 @@ const ExplorerSearchResultsTable = ({
...
-    const [numResults] = useState(data.length);
+    const [activeFilters, setActiveFilters] = useState([]);
 
     const [summaryModalVisible, setSummaryModalVisible] = useState(false);
     const [tracksModalVisible, setTracksModalVisible] = useState(false);
 
+    const filteredData = useMemo(() => {
+        return data.filter((item) => {
+            return Object.keys(activeFilters).every((filterKey) => {
+                const filterValues = activeFilters[filterKey] || [];
```
davidlougheed, line 40: object.keys + every + a prop access is a confusing way to write this, when it could just be `Object.entries(activeFilters).every(([filterKey, filterValues]) => ...)`. Also, I'd prefer expression-based functions over the multiple-nested-closure with return style seen here.

src/components/explorer/ExplorerSearchResultsTable.js
```diff
@@ -29,11 +29,25 @@ const ExplorerSearchResultsTable = ({
...
+                if (filterValues.length === 0) {
+                    return true;
+                }
+                return filterValues.includes(item[filterKey]);
+            });
+        });
+    }, [data, activeFilters]);
+
+    const numResults = useMemo(() => filteredData.length, [filteredData]);
```
davidlougheed, line 49: this memo doesn't do much, since `.length` is a prop of filteredData that doesn't take much time to compute and isn't an object. I'd remove this.

src/components/explorer/searchResultsTables/ExperimentsTable.js
```diff
@@ -28,59 +28,67 @@ ExperimentRender.propTypes = {
...
-
 const ExperimentsTable = ({ data, datasetID }) => {
     const tableSortOrder = useSelector(
         (state) => state.explorer.tableSortOrderByDatasetID[datasetID]?.["experiments"],
     );
 
+    const experimentTypeFilters = useDynamicTableFilterOptions(data, "experimentType");
+
+    const columns = useMemo(() => {
+        const baseColumns = [
```
davidlougheed, line 39: just return this array directly instead of assigning it to a variable which immediately gets returned.

davidlougheed (changes requested):
when I filter and then export CSV without checking any checkbox, I expect to get only the rows I filtered back. instead, I get the whole of the search results back.

davidlougheed (changes requested):

davidlougheed:
otherwise mostly working well!

davidlougheed (approved):
lgtm

PR #409, opened 2024-06-19
Unified eslint and prettier

SanjeevLakhwani:
- made eslint and prettier be in harmony with each other
- ran prettier and lint checked


package.json
```diff
@@ -66,13 +68,14 @@
     "worker-loader": "^3.0.8"
   },
   "scripts": {
-    "lint": "npx eslint src/",
-    "lint-fix": "npx eslint src/ --fix",
+    "lint": "eslint 'src/**/*.{js,jsx,ts,tsx}'",
+    "lint-fix": "eslint 'src/**/*.{js,jsx,ts,tsx}' --fix",
+    "prettier": "prettier --write 'src/**/*.{js,jsx,ts,tsx}'",
```
v-rocheleau, line 73: We should have a `prettier` script that simply checks the prettier rules, and a `prettier-fix` that actually writes the changes. We should also add a prettier step to the CI lint job.

SanjeevLakhwani: the prettier/prettier error rule in eslint checks for it, therefore we do not need a prettier-fix command

v-rocheleau (approved):
Lgtm

PR #414, opened 2024-06-26
add text search to drop-box

noctillion:


src/components/manager/ManagerDropBoxContent.js
```diff
@@ -20,6 +20,7 @@ import {
   Tree,
   Typography,
   Upload,
+  Input,
```
davidlougheed, line 23: alphabet order

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -293,21 +294,44 @@ const ManagerDropBoxContent = () => {
...
+
   const filesByPath = useMemo(
     () => Object.fromEntries(recursivelyFlattenFileTree([], tree).map((f) => [f.relativePath, f])),
     [tree],
   );
 
-  const treeData = useMemo(
-    () => [
+  const filterTree = (nodes) => {
+    return nodes.reduce((acc, node) => {
```
davidlougheed, line 309: this should be a filter, not a reduce I think

davidlougheed: also, I think this function should be moved outside of the component, since it isn't a closure.

davidlougheed (changes requested):
initial pass

src/components/manager/ManagerDropBoxContent.js
noctillion: I am not sure about filter,  could reduce be better here because the hierarchical structure of the folders?

davidlougheed: ah yes i see, its more of a filter-map. this works then.

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -293,21 +311,27 @@ const ManagerDropBoxContent = () => {
   const ingestionWorkflows = workflowsByType.ingestion.items;
   const ingestionWorkflowsByID = workflowsByType.ingestion.itemsByID;
 
+  const [searchTerm, setSearchTerm] = useState("");
+
+  const handleSearchChange = (event) => {
```
davidlougheed, line 316: wrap in useCallback

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -572,6 +596,13 @@ const ManagerDropBoxContent = () => {
             <Button icon={<UploadOutlined />} onClick={handleUpload} disabled={uploadDisabled}>
               Upload
             </Button>
+
+            <Input
+              placeholder="Search files..."
+              value={searchTerm}
+              onChange={handleSearchChange}
+              style={{ marginBottom: 8, maxWidth: 300 }}
+            />
```
davidlougheed, line 605: add extra code newline after for spacing

src/components/manager/ManagerDropBoxContent.js
```diff
@@ -572,6 +596,13 @@ const ManagerDropBoxContent = () => {
             <Button icon={<UploadOutlined />} onClick={handleUpload} disabled={uploadDisabled}>
               Upload
             </Button>
+
+            <Input
```
davidlougheed, line 600: add `allowClear={true}` to allow users to clear search in one click

davidlougheed (changes requested):

davidlougheed (approved):
lgtm

PR #415, opened 2024-07-09
chore: update react-pdf + rewrite PdfDisplay as TS

davidlougheed:


v-rocheleau (approved):
Looks/behaves good to me. Double checked the release notes for breaking changes since we bump 2 major versions, all in order.

PR #416, opened 2024-07-09
refact: rewrite file/JSON display components as TypeScript

davidlougheed:


src/components/display/CsvDisplay.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+            let innerParseError: string | null = null;
+
+            // noinspection JSUnusedGlobalSymbols
+            Papa.parse<string[]>(csvText, {
+              worker: true,
+              step: (res) => {
+                if (res.errors?.length) {
+                  innerParseError = res.errors[0].message;
+                }
```
v-rocheleau, line 45: (very) minor: in case of an error on an early row, does it keep parsing rows until the end of the document?
If so, it would be better to have an early reject on the first error encountered, since we only show an alert in that case.

src/components/display/DocxDisplay.tsx
```diff
@@ -12,17 +11,22 @@ const MAMMOTH_OPTIONS = {
...
+const styles: Record<string, CSSProperties> = {
   container: {
     maxWidth: 960, // Maximum width to roughly a nice page
     overflowX: "auto",
   },
 };
 
-const DocxDisplay = ({ contents, loading }) => {
+type DocxDisplayProps = {
+  contents: Blob;
```
v-rocheleau, line 22: `contents?: Blob;` instead?
We could also have a reusable generic props type for file content display, since we are using the same props in the other components.

src/components/display/CsvDisplay.tsx
```diff
@@ -0,0 +1,89 @@
+import { useEffect, useState } from "react";
+import Papa from "papaparse";
+import { Alert } from "antd";
+
+import SpreadsheetTable, { SPREADSHEET_ROW_KEY_PROP, SpreadsheetTableProps } from "./SpreadsheetTable";
+
+const DEFAULT_COLUMN = { key: "col" };
+
+type CsvDisplayProps = {
+  contents?: Blob | null;
```
v-rocheleau, line 10: I dont think the null type necessary here.

src/components/display/CsvDisplay.tsx
```diff
@@ -0,0 +1,89 @@
...
+type CsvParseResult = [SpreadsheetTableProps<CsvRecord>["columns"], CsvData];
+
+const CsvDisplay = ({ contents, loading }: CsvDisplayProps) => {
+  const [parsedData, setParsedData] = useState<CsvData>([]);
+  const [parseError, setParseError] = useState("");
+  const [isParsing, setIsParsing] = useState(true); // Start in parsing state
+  const [columns, setColumns] = useState<SpreadsheetTableProps<CsvRecord>["columns"]>([DEFAULT_COLUMN]);
+
+  useEffect(() => {
+    if (contents === undefined || contents === null) return;
```
v-rocheleau, line 26: `if (!contents) return` ?

v-rocheleau (changes requested):
first pass

src/components/display/CsvDisplay.tsx
davidlougheed: since the parsing isn't done as a promise, early reject may mean it keeps parsing after promise rejection, which might get messy. I'll look into the best thing to do here.

davidlougheed: for now I just aggregate multiple errors as it goes along

gsfk (approved):

src/components/display/CsvDisplay.tsx
```diff
@@ -0,0 +1,85 @@
+import { useEffect, useState } from "react";
+import Papa from "papaparse";
+import { Alert } from "antd";
+
+import SpreadsheetTable, { SPREADSHEET_ROW_KEY_PROP, SpreadsheetTableProps } from "./SpreadsheetTable";
```
davidlougheed, line 5: should be `type SpreadsheetTableProps`

PR #417, opened 2024-07-16
lint: check for hook misusage in lint step [redo] 

davidlougheed:


SanjeevLakhwani (approved):
Looks good, appreciate the effort!

gsfk (commented):
Only hurdle for me here is that `DiscoverySearchForm.js` appears to be broken in current master, so hard to test changes.

PR #419, opened 2024-07-18
fix: project json schemas types query in a hook

v-rocheleau:
Removes `fetchServiceDependentData`, which was only used to fetch extra prop schema types, and replaces it with a hook.

### DIFF
--- a/src/components/manager/projects/ProjectJsonSchemaModal.js
+++ b/src/components/manager/projects/ProjectJsonSchemaModal.js
@@ -6,17 +6,15 @@ import { Button, Form, Modal } from "antd";
 import { PlusOutlined } from "@ant-design/icons";
 
 import { createProjectJsonSchema } from "@/modules/metadata/actions";
-import ProjectJsonSchemaForm from "./ProjectJsonSchemaForm";
 import { useProjectJsonSchemaTypes } from "@/modules/metadata/hooks";
+
+import ProjectJsonSchemaForm from "./ProjectJsonSchemaForm";
 
 const ProjectJsonSchemaModal = ({ projectId, open, onOk, onCancel }) => {
   const dispatch = useDispatch();
 
-  const {
-    isFetching: isFetchingExtraPropertiesSchemaTypes,
-    isCreating: isCreatingJsonSchema,
-    schemaTypes: extraPropertiesSchemaTypes,
-  } = useProjectJsonSchemaTypes();
+  const { isFetchingExtraPropertiesSchemaTypes, isCreatingJsonSchema, extraPropertiesSchemaTypes } =
+    useProjectJsonSchemaTypes();
 
   const [form] = Form.useForm();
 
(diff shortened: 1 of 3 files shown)


src/modules/metadata/hooks.js
```diff
@@ -33,3 +33,16 @@ export const useOverviewSummary = () => {
...
   return useSelector((state) => state.overviewSummary);
 };
+
+export const useProjectJsonSchemaTypes = () => {
+  const dispatch = useDispatch();
+  const metadataService = useService("metadata");
+  useEffect(() => {
+    dispatch(fetchExtraPropertiesSchemaTypes());
+  }, [dispatch, metadataService]);
+  return useSelector((state) => ({
```
davidlougheed, line 43: selector shouldn't return an object, since it'll change every time. instead, there should be a deconstruction of `useProjects()` to the three properties, and then a returned memoized object of the three properties

src/modules/user/actions.js
```diff
@@ -27,7 +24,7 @@ export const fetchUserDependentData = (servicesCb) => async (dispatch, getState)
       // If we're newly authenticated as an owner, we run all actions that may have changed with authentication
       // (via the callback).
       // TODO: invalidate projects/datasets/other user-dependent data
-      await dispatch(fetchServicesWithMetadataAndDataTypesIfNeeded(() => dispatch(fetchServiceDependentData())));
+      await dispatch(fetchServicesWithMetadataAndDataTypesIfNeeded());
```
davidlougheed, line 27: :)))

src/components/manager/projects/ProjectJsonSchemaModal.js
```diff
@@ -1,21 +1,22 @@
...
-import { useDispatch, useSelector } from "react-redux";
+import { useDispatch } from "react-redux";
 import PropTypes from "prop-types";
 
 import { Button, Form, Modal } from "antd";
 import { PlusOutlined } from "@ant-design/icons";
 
 import { createProjectJsonSchema } from "@/modules/metadata/actions";
 import ProjectJsonSchemaForm from "./ProjectJsonSchemaForm";
+import { useProjectJsonSchemaTypes } from "@/modules/metadata/hooks";
```
davidlougheed, line 10: group with other modules/metadata import

src/modules/metadata/hooks.js
```diff
@@ -33,3 +33,16 @@ export const useOverviewSummary = () => {
   }, [dispatch, metadataService]);
   return useSelector((state) => state.overviewSummary);
 };
+
+export const useProjectJsonSchemaTypes = () => {
+  const dispatch = useDispatch();
+  const metadataService = useService("metadata");
+  useEffect(() => {
+    dispatch(fetchExtraPropertiesSchemaTypes());
```
davidlougheed, line 41: does this action properly check if the metadata service exists before trying to execute?

davidlougheed (changes requested):

src/modules/metadata/hooks.js
v-rocheleau: Yes it does

davidlougheed (approved):
Lgtm

PR #422, opened 2024-07-30
Features/advanced search tables

noctillion:
add experiments tab for advanced search results

### DIFF
--- a/src/modules/explorer/reducers.js
+++ b/src/modules/explorer/reducers.js
@@ -27,7 +27,9 @@ import {
   SET_OTHER_THRESHOLD_PERCENTAGE,
   SET_IGV_POSITION,
   FETCH_IGV_GENOMES,
+  CLEAR_SEARCH,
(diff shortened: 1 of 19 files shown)


src/modules/explorer/reducers.js
```diff
@@ -291,7 +291,7 @@ export const explorer = (
           [action.datasetID]: {
             results: freeTextResults(action.data),
             searchFormattedResults: tableSearchResults(action.data),
-            searchFormattedResultsExperiments: tableSearchResultsExperiments(action.data),
+            searchFormattedResultsExperiment: tableSearchResultsExperiments(action.data),
```
davidlougheed, line 294: why this change?

davidlougheed: i would expect plural but it is singular elsewhere i guess?

davidlougheed (approved):
lgtm

PR #427, opened 2024-08-22
refact: open grafana in new tab and user role check for display

CowTrainer:


v-rocheleau:
Grafana now opens in a new tab, rather than in an iframe embedded in Bento.
The Grafana button render is now conditional on the user's roles, which is determined by the ID token's content.

davidlougheed (approved):
lgtm

PR #429, opened 2024-08-26
fix(explorer): fix variant querying not working

davidlougheed:


src/components/discovery/DiscoverySearchForm.js
```diff
@@ -250,14 +250,15 @@ const DiscoverySearchForm = ({ onChange, dataType, setFormRef, handleVariantHidd
         updatedConditionsArray = updateVariantConditions(updatedConditionsArray, "[dataset item].alternative", alt);
       }
 
+      form.setFieldsValue({ conditions: updatedConditionsArray }); // update form separately - not controlled
```
davidlougheed, line 253: this is the actual fix

PR #432, opened 2024-08-28
refact: add typing in some straightforward places

davidlougheed:


gsfk (approved):
is it worth deduplicating `Record<string, CSSProperties>` at some point?

davidlougheed:
yes, I think that would be good

PR #437, opened 2024-09-09
refact: new drop box JSON select form input + fixed JSON type

davidlougheed:
trying here to address some of the visual hierarchy issues with the `DropBoxJsonSelect` component, plus splitting/reorganizing the files between it and `DropBoxTreeSelect`. This also converts `DropBoxJsonSelect` into a real form input, usable inside a `Form.Item`.

also fixes a bad JSON type for typescript.


src/components/manager/dropBox/DropBoxJsonSelect.tsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+export type DropBoxJsonSelectProps = {
+  initialValue?: JSONType;
+  value?: JSONType;
+  onChange?: (x: JSONType) => void;
+};
+
+/** A form input component for DropBox JSON file selection. */
+const DropBoxJsonSelect = ({ initialValue, onChange }: DropBoxJsonSelectProps) => {
```
v-rocheleau, line 38: Way better, thanks! I think I might add a radio option for "None", since Discovery config is optional.
The user can submit an empty discovery config by selecting nothing, but I don't think it is clear enough to the user.

v-rocheleau (approved):

src/components/manager/dropBox/DropBoxJsonSelect.tsx
davidlougheed: that's a good idea!

PR #444, opened 2024-09-11
chore: remove a check for the old Grafana iframe tab

davidlougheed:


v-rocheleau (approved):
oops forgot that, lgtm!

PR #448, opened 2024-09-18
refact: rewrite App as TSX + remove dead pingInterval code

davidlougheed:
also updates bento-auth-js to 6.0.2


src/components/App.tsx
```diff
@@ -197,14 +190,13 @@ const App = () => {
   return (
     <ChartConfigProvider
       Lng="en"
-      translationMap={{ en: { Count: "Count", Other: "Other" } }}
+      translationMap={{ en: { Count: "Count", Other: "Other" }, fr: { Count: "Observations", Other: "Autre" } }}
```
gsfk, line 193: The French "Obervations" is different from [the bento_charts default](https://github.com/bento-platform/bento_charts/blob/main/src/constants/chartConstants.ts#L14-L17) "Comptage", the term currently used on French instances of bento public:

```
  fr: {
    [COUNT_KEY]: 'Comptage',
    [OTHER_KEY]: 'Autre',
  },
```
I see that the bento public (not bento charts) [default translation](https://github.com/bento-platform/bento_public/blob/main/src/public/locales/fr/default_translation_fr.json) has these lines: 

```
  "Count": "Observations",
  "Counts": "Totaux",
```

... but these don't seem to be used anywhere.

gsfk (commented):
minor quibble about translation, otherwise good

src/components/App.tsx
davidlougheed: yeah i was going based on public, but it doesn't look right to me either. wonder where it came from... some errant artificial 'intelligence' somewhere perhaps

gsfk: it may have taken us a while to decide, and a lot of cruft was left behind

gsfk (approved):

PR #449, opened 2024-09-18
feat: show a message when the reference service is down

davidlougheed:


v-rocheleau (approved):
Lgtm, good idea to have a ServiceError component

PR #450, opened 2024-09-19
refact: rewrite service detail pages as .tsx + show not found

davidlougheed:
service detail pages are now in typescript.

if a bad service (or a service which is down) is specified, an error is now shown:

image


src/components/services/ServiceDetail.tsx
v-rocheleau, line 1: Minor: If I set a service kind that shouldn't exist manually in the URL (e.g. `services/dummy/overview`), this component will display an error result telling us that the `dummy` service is down.
Normally this situation should not happen when a user navigates in web by clicking around, but I think it is worth noting.

image

davidlougheed: altered the wording a bit

v-rocheleau (approved):
lgtm

PR #454, opened 2024-09-20
chore!: remove service itemsByArtifact + rm old fallback code

davidlougheed:


src/modules/services/reducers.js
```diff
@@ -87,18 +70,16 @@ export const services = (
...
 
       return {
         ...state,
 
         items,
         itemsByID,
         itemsByKind,
-        itemsByArtifact,
 
         // Backwards-compatibility with older Bento versions, where this was called 'federation'
```
v-rocheleau, line 81: This comment no longer applies with these changes, right?

davidlougheed: yes, oops

v-rocheleau (approved):
Lgtm

PR #455, opened 2024-09-20
fix: dropbox form select for nested files

v-rocheleau:
Dropbox form selector was only matching file paths at the root, preventing the selection of files nested in a directory.
Now recursively searches inside the directories.

### DIFF
--- a/src/modules/dropBox/hooks.ts
+++ b/src/modules/dropBox/hooks.ts
@@ -7,6 +7,7 @@ import { type RootState, useAppDispatch, useAppSelector } from "@/store";
 import type { JSONType } from "@/types/json";
 
 import { fetchDropBoxTree } from "./actions";
+import type { DropBoxEntry } from "./types";
 
 export const useDropBox = () => {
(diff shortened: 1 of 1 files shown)


src/modules/dropBox/hooks.ts
```diff
@@ -24,19 +25,46 @@ export const useDropBox = () => {
...
+      if (result) {
+        return result;
+      }
+    }
+  }
+  return null;
+};
+
+const useFindDropBoxEntry = (filePath?: string): DropBoxEntry | null => {
+  const tree: DropBoxEntry[] = useAppSelector((state: RootState) => state.dropBox.tree);
```
davidlougheed, line 43: this should be `useDropBox` instead of a selector i think

davidlougheed:
```ts
const tree: DropBoxEntry[] = useDropBox().tree;
```

davidlougheed (changes requested):

davidlougheed (approved):
lgtm!

PR #458, opened 2024-10-25
fix: error boundary for data type summaries + fix variant crash

davidlougheed:
also refactors dataset summary redux code to fix a bunch of logic errors contained within.

Redmine: https://redmine.c3g-app.sd4h.ca/issues/2149


gsfk (commented):
Seems to work fine, variant not crashing, although the "VCF Files" part of the summary never appears. Perhaps that's beyond the scope of this PR. 

Other comments:

- The error boundary is nice, although seems super-specific to this one case ~~where there's actually very little code to get wrong~~ Do we want a more general error boundary higher up in the code? Again this may be an out of scope comment.
- I'm mostly indifferent to whether we use `datasetId` or `datasetID`, but this pr has both (I see datasetId in Dataset.js).

davidlougheed:
re: the error boundary, ive seen crashes come from this part of the code an oddly high amount of times – we do eventually want error boundaries in other places, but this is a convenient place to put one for now.

davidlougheed:
regarding ID capitalization, it's inconsistent everywhere but i've reverted a diff where i changed one existing case.

gsfk:
> re: the error boundary, ive seen crashes come from this part of the code an oddly high amount of times – we do eventually want error boundaries in other places, but this is a convenient place to put one for now.

Yeah, I wrote that comment before I realized that failing API calls were behind a lot of the problems, I think this makes sense here.

gsfk (approved):

PR #459, opened 2024-10-28
feat: HTML file display support

davidlougheed:


src/components/display/HtmlDisplay.tsx
```diff
@@ -0,0 +1,76 @@
+import { useEffect, useId, useState } from "react";
+import { Spin } from "antd";
+
+import type { BlobDisplayProps } from "./types";
+import { LoadingOutlined } from "@ant-design/icons";
+
+const HtmlDisplay = ({ contents, loading }: BlobDisplayProps) => {
+  const iframeId = useId();
```
noctillion, line 8: Not sure if it is worth using useRef instead?

noctillion (approved):
It works well. Lgtm

src/components/display/HtmlDisplay.tsx
davidlougheed: good point

PR #465, opened 2024-11-11
fix: workflow boolean inputs default value

v-rocheleau:
Workflow inputs of type "boolean" don't have their default value set correctly in the antd Form.
If the input's initial value is not present in the `initialValues` prop, the value defaults to `undefined`.

This bug causes errors when ingesting VCFs, "filter out references" can only be `false` by checking and unchecking the input.
, causing WES to reject the request due to a missing workflow input.
[Ticket](https://redmine.c3g-app.sd4h.ca/issues/2390)


davidlougheed (approved):
lgtm

PR #466, opened 2024-11-12
refact: VariantSearchHeader types + locus search fixes

davidlougheed:
should fix issues when "searching" for genes without having any GFF3 file ingested in the reference genome.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -49,6 +49,7 @@
         "@babel/core": "^7.25.2",
         "@babel/preset-env": "^7.25.4",
         "@babel/preset-react": "^7.24.1",
+        "@types/json-schema": "^7.0.15",
         "@types/papaparse": "^5.3.14",
         "@types/react": "^18.3.5",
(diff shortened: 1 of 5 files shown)


gsfk (commented):
Seems to work correctly, with a couple comments:

- a few times I got this antd warning: 
```DiscoverySearchForm.js:149 Warning: Instance created by `useForm` is not connected to any Form element. Forget to pass `form` prop?```

-  a stale "position" is sometimes passed in the form despite the field in the UI being blank

More general comments about the variant search UI... not specific to this pr, but writing down anyhow: 

- The fields in "Assembly ID" are pulled from gohan, and may or may not match the ids in the reference service, so gene search can break if there's a mismatch

- The list of assemblies is not scoped at all, so you are always offered all options no matter what project or dataset you choose.

gsfk:
works, although now allows empty fields to launch queries (instead of expected behaviour of marking empty field in red..... )

gsfk (approved):

PR #468, opened 2024-11-15
fix: proper notification mark-as-read behaviour

davidlougheed:


gsfk (approved):
Number unread dropped by hundreds as soon as I loaded the branch, lolz. Works well.

PR #474, opened 2024-12-18
feat(explorer): search row expansion with details

davidlougheed:


gsfk (commented):
Actually took me a while to figure out what the functionality was, seems to work as intended. I can't tell if I'd ever use it in lieu of clicking the entry. 

ExperimentsTable.js gives me a React warning, although it seems like it the fault of the existing code. Specifically this line: 
```render: (experimentId, record) => <ExperimentRender experimentId={experimentId} {...record} />```

gives this warning about spread: 

```ExperimentsTable.js:73 Warning: A props object containing a "key" prop is being spread into JSX. React keys must be passed directly to JSX without using spread.```

davidlougheed:
most of the point of this is to make experiment results files accessible in fewer clicks; the other expansions I just added because it was easy to do so. I'll look into fixing the key error...

gsfk (approved):

PR #475, opened 2025-01-06
Fixes/variants form submit

gsfk:
Fixes issues with search form validation, where users were permitted to launch searches with invalid form state. (Redmine [2415](https://redmine.c3g-app.sd4h.ca/issues/2415)). 

Some of the previous tricks for variants form validation  do not work as well in antd > 3, this is fixed by moving validation code to a parent component and invoking manually when needed. 

Also discovered and fixed this issue of uncertain age: data types are not removed from the submitted form when deleting data type tabs (when removing, e.g. the "Variants" search tab, the submitted form still has fields for variants). This can block form submission, and without giving any user feedback (because the tab is not rendered).

### DIFF
--- a/src/components/discovery/DiscoveryQueryBuilder.js
+++ b/src/components/discovery/DiscoveryQueryBuilder.js
@@ -71,7 +71,7 @@ const DiscoveryQueryBuilder = ({ activeDataset, dataTypeForms, requiredDataTypes
       if (action !== "remove") return;
       dispatch(removeDataTypeQueryForm(activeDataset, dataTypesByID[key]));
 
-      // remove this field from antd form
+      // remove this field from submitted form
       setForms((fs) => objectWithoutProp(fs, key));
     },
     [dispatch, activeDataset, dataTypesByID],
--- a/src/components/discovery/DiscoveryQueryBuilder.js
+++ b/src/components/discovery/DiscoveryQueryBuilder.js
@@ -14,9 +14,36 @@ import {
 import { useDataTypes, useServices } from "@/modules/services/hooks";
 import { useAppDispatch, useAppSelector } from "@/store";
 import { nop } from "@/utils/misc";
+import { VARIANT_OPTIONAL_FIELDS } from "@/utils/search";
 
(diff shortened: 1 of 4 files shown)


src/components/discovery/DiscoverySearchForm.js
```diff
@@ -92,9 +55,17 @@ PhenopacketDropdownOption.propTypes = {
...
+const DiscoverySearchForm = ({
+  onChange,
+  dataType,
+  setFormRef,
+  handleVariantHiddenFieldChange,
+  conditionValidator,
+}) => {
   const [form] = Form.useForm();
 
+  const CONDITION_RULES = [{ validator: conditionValidator }];
```
davidlougheed, line 67: since this is inside a function, it shouldn't be all-caps constant-named anymore - just a normal variable

davidlougheed: or, if conditionValidator is treated as an import instead, this could stay a constant outside of the component function.

src/components/discovery/DiscoveryQueryBuilder.js
```diff
@@ -133,6 +169,7 @@ const DiscoveryQueryBuilder = ({ activeDataset, dataTypeForms, requiredDataTypes
               setFormRef={setFormRef}
               onChange={handleChange}
               handleVariantHiddenFieldChange={handleVariantHiddenFieldChange}
+              conditionValidator={conditionValidator}
```
davidlougheed, line 172: since this is always the same value, I think it should be moved to utils/search instead of being passed as a prop, and then imported for both uses.

davidlougheed (changes requested):

src/components/discovery/DiscoverySearchForm.js
davidlougheed: @gsfk can you move this back outside of DiscoverySearchForm to avoid redefining it every render?

gsfk: ah, managed to overlook this while moving everything around

davidlougheed (approved):
lgtm, thanks!

PR #480, opened 2025-02-13
refact: add PermissionsGate + debug state/log for forbidden errors

davidlougheed:


gsfk (approved):
Seems to work as intended. 

Only issues I could find are all present in current master (for example the red X "Forbidden" message doesn't show when trying to see an individual without permissions, you just get a temporary notification and the antd skeleton).

gsfk:
Managed to get one weird flash on private Firefox (saw Datasets & Projects for 1/4 second before "Forbidden" kicked in.... but possibly just cache from having been logged in with full permissions moments before?)

PR #483, opened 2025-04-01
chore: include credentials for projects list requests

v-rocheleau:
Include credentials in requests to obtain the projects list.


src/modules/metadata/actions.js
```diff
@@ -62,7 +62,7 @@ export const clearDatasetDataType = networkAction((datasetId, dataTypeID) => (_d
 const fetchProjects = networkAction(() => (_dispatch, getState) => ({
   types: FETCH_PROJECTS,
   url: `${getState().services.metadataService.url}/api/projects`,
-  publicEndpoint: true,
+  publicEndpoint: false,
```
davidlougheed, line 65: this defaults to false-y, so i'd just cut this line out. I'm fine with merging this before the rest, since there's no harm in passing the token to katsu even if it goes unused for now.

PR #484, opened 2025-04-11
fix: catch error when discovery_config is not a valid JSON

noctillion:
catch error if config file is not a valid json preventing rendering failure


src/modules/dropBox/hooks.ts
```diff
@@ -130,6 +130,13 @@ export const useDropBoxJsonContent = (filePath?: string, defaultValue: JSONType
...
     }
     // parsed file content, or null if empty
-    return rawText ? JSON.parse(rawText) : null;
+    try {
+      return rawText ? JSON.parse(rawText) : null;
+    } catch (e) {
+      // if the file is not a valid JSON
+      return {
+        error: `${e}`,
+      };
```
v-rocheleau, line 139: If there is no parsing errors this function returns a JSON.
If we also return a JSON when there is a parsing error, the user may think that what they get here is the actual content of the file.
One thing we could do is have this hook return 2 things: the JSON content, and an error message if any occurs.

v-rocheleau (changes requested):

v-rocheleau (approved):
Tentative approval.

We no longer have uncaught errors causing issues in the projects/datasets forms, but the error display in the form should get more considerations.

For instance, when trying to use an experiments object as a config, I get repeated error messages:
image

PR #489, opened 2025-06-09
fix: JSON schema validation

davidlougheed:


davidlougheed:
but an array is indeed not a valid discovery schema, wonder what happened here then

gsfk:
Appears to be a katsu issue, `edge` is returning discovery config as an array.

PR #495, opened 2025-08-13
refact: rewrite DataManagerContent and SiteHeader as TSX

davidlougheed:


src/components/DataManagerContent.tsx
```diff
@@ -23,8 +24,8 @@ const styles = {
     marginLeft: "-24px",
     marginRight: "-24px",
     marginTop: "-12px",
-  },
-  suspenseFallback: { padding: "24px", backgroundColor: "white" },
+  } as CSSProperties,
```
SanjeevLakhwani, line 27: Might be better to type the object instead. `Record<string, CSSProperties>`

SanjeevLakhwani (commented):
lint

src/components/DataManagerContent.tsx
davidlougheed: reason I constructed it like this was because my IDE doesn't like to autocomplete properties if it thinks there can be any string as a property. not ideal tho

SanjeevLakhwani (approved):
lgtm

PR #498, opened 2025-09-08
feat(explorer): display instrument device / ontology

noctillion:
Render device and device ontology in experiments — aligns with Features/instrument ontologies #563 in katsu

### DIFF
--- a/src/components/explorer/IndividualExperiments.js
+++ b/src/components/explorer/IndividualExperiments.js
@@ -6,7 +6,12 @@ import { Button, Descriptions, Popover, Table, Tooltip, Typography } from "antd"
 import { BarsOutlined, EyeOutlined, FileTextOutlined, ProfileOutlined } from "@ant-design/icons";
 
 import { EM_DASH } from "@/constants";
-import { experimentPropTypesShape, experimentResultPropTypesShape, individualPropTypesShape } from "@/propTypes";
+import {
+  experimentPropTypesShape,
+  experimentResultPropTypesShape,
+  individualPropTypesShape,
+  ontologyShape,
+} from "@/propTypes";
 import { getFileDownloadUrlsFromDrs } from "@/modules/drs/actions";
 import { useAppDispatch, useAppSelector } from "@/store";
 import { guessFileType } from "@/utils/files";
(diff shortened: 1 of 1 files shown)


src/components/explorer/IndividualExperiments.js
```diff
@@ -206,15 +206,11 @@ export const ExperimentDetail = ({ experiment }) => {
...
-            </div>
-            <div>
-              <strong>ID:</strong>&nbsp;<MonospaceText>{instrument.identifier}</MonospaceText>
-            </div>
-          </div>
+        <Descriptions.Item span={1} label="Device">
+          {instrument.identifier}
+        </Descriptions.Item>
+        <Descriptions.Item span={1} label="Device Ontology">
+          <OntologyTerm term={instrument.device_ontology?.[0] ?? instrument.device_ontology} />
```
v-rocheleau, line 213: To link with this [discussion](https://github.com/bento-platform/katsu/pull/563/files#r2242808011) on the associated Katsu PR:

If we support multiple ontologies for instruments in the backend, we should do so in the frontend as well.
This only displays a single ontology.

src/components/explorer/IndividualExperiments.js
```diff
@@ -206,15 +206,11 @@ export const ExperimentDetail = ({ experiment }) => {
...
-          <div style={{ display: "flex", gap: 16 }}>
-            <div>
-              <strong>Platform:</strong>&nbsp;{instrument.platform}
-            </div>
-            <div>
-              <strong>ID:</strong>&nbsp;<MonospaceText>{instrument.identifier}</MonospaceText>
-            </div>
-          </div>
+        <Descriptions.Item span={1} label="Device">
+          {instrument.identifier}
```
davidlougheed, line 210: q: meant to be `.device`?

davidlougheed: *suggestion(non-blocking):* This also now leaves out instrument ID (which maybe is a good thing). To me, it would make sense to keep the original instrument description with nested divs perhaps, + the now have Device be `instrument.device (<OntologyTerm term={instrument.device_ontology} />)` with conditional rendering of the ontology term if it exists, in parentheses, to link the concept of a free-text device name and the corresponding ontology term.

v-rocheleau: Went ahead with your suggestion

PR #514, opened 2026-01-12
feat: support creating drop box subfolders in upload

davidlougheed:
also updates to antd 5.29
see newly-fleshed-out ticket https://redmine.c3g-app.sd4h.ca/issues/1949


gsfk (commented):
A few mostly non-blocking comments: 

- shouldn't it block upload if you enter an illegal subfolder name? (currently it will do the upload, but pretend you didn't create a subfolder)
- it seems like there is some stale state going on with the parent folder, because I get a stale parent folder selection this way:
  - click a folder in the main Drop Box UI and click "Upload". The parent folder in the modal will be correct. 
  - close the modal (either by finishing upload or just by closing). If you now click a different folder in the main UI and click "Upload" again the parent folder in the modal will be wrong: it will be your choice from last time. If you choose & click multiple times it will always be one selection behind. 
  
Not changed in this PR but I was confused by the File Selector in the Upload modal, since the button to select a file is called "Upload". Shouldn't it be "Select File"? Presumably the final "OK" button means upload.

gsfk (approved):

src/components/manager/dropBox/DropBoxTreeSelect.tsx
```diff
@@ -10,64 +13,329 @@ import { sortByName } from "./common";
...
+
+  return (
+    <Space.Compact>
+      {currentPath ? (
+        <Space.Addon style={{ whiteSpace: "nowrap" }}>{currentPath + (currentPath === "/" ? "" : "/")}</Space.Addon>
+      ) : null}
+      <Input
+        value={newSubfolderName}
+        onChange={(e) => {
+          setNewSubfolderName(e.target.value.replaceAll(/[\s/*]/g, "_"));
```
SanjeevLakhwani: the backward slash is also an illegal character on linux. Also should we disable `.` just in case. I dont see how someone could use this for now, but to be safe.

SanjeevLakhwani (changes requested):
First pass.

The UI for creating a new subfolder could use some improvement. Ideally I would prefer a new folder icon in row with the folder in the tree and also an option in the following: 

Image


Also, if we are allowing adding subfolders we should also have a way to delete them.

PR #534, opened 2026-03-23
feat: dataset provenance form with view/edit/validation

SanjeevLakhwani:
## Summary

- Adds a new `DatasetForm` with tabbed UI covering all dataset metadata fields (core info, contacts, links, classification, study details, publications/funding, PCGL info)
- Adds a **View Provenance** button on the Dataset card (visible to all users) that opens a read-only modal of all dataset metadata
- Migrates all `project.datasets` references to `project.datasets_v2` and removes the deprecated `datasets` field from the `Project` type

## Pre-requisite
- Requires: https://github.com/bento-platform/katsu/pull/665
- Requires latest gohan


gsfk:
First comment: wipes out existing datasets in web, although public seems fine. 

Or perhaps I'm meant to migrate? It's not clear how to do that, however.

gsfk (commented):
First-pass attempt, I haven't tried everything yet. 

- Existing datasets go missing, see comment above.
- Overall, the form is pretty long, so I suspect will be easier to submit as a document. I haven't tried this yet
- Because it's so long it should be a little more forgiving, in particular a mis-click outside of the modal should not close the modal and wipe all content.
- Does it need to be a modal?
- The "person" and "organization" roles are all together ("Site" doesn't make much sense as a role for a person)
- it wasn't immediately clear to me that "Primary Contact" had to be a single entity... I guess that's on me for not reading the word "primary", although possibly I was put off by "add name" which looks like you can add more contacts. 
- "add Ontology resource" is bound to be filled in incorrectly without more controls, even in day-to-day bento operations you can find data with incorrect iri_prefix. This is more of a bento bookkeeping comment than a review of this particular PR. It refuses ontology additions without resources, so not sure if this will discourage ontology use rather than encourage it. I'm not sure how to make this easier. 
- Validation errors produce the message "Zod validation error" which is a little odd sounding
- Are stakeholders really required? What if the primary contact is an organization and also the stakeholder?

- after all that, it actually failed to create a dataset, since katsu seems to require `release_date` but the form doesn't, so it failed (and I lost everything I typed in...)

gsfk (changes requested):
Coming along!

- can't actually create a dataset, even when form passes validation, although that now looks like a katsu error. It produces `400 Bad Request` and this message: "`Error validating discovery scope: project-dataset pair does not exist."` which seems like a nonsensical error when trying to create a dataset
- some fields are required dynamically (license url is required only if some other license fields are filled in) but are never marked as required in the form
- I'm able to fill in keywords using both string and ontology even though the form says it should be one or the other
- submission by file seems to work well, although can't test because of first issue


less important stuff:
- validation errors could show which section of the form they're from (currently they show intenal form codes like `root` or` taxa.0`)
- validation errors are sometimes incomplete: if I add ontologies without a resource and a license without a url, it only shows the license error. If I correct the license, the ontology error appears

gsfk (changes requested):
Validation errors are missing... now only shows errors on the current tab, if any. 

If you don't have errors on the current tab it just blocks submission without telling you anything.

gsfk (approved):
working as expected, pending migration away from "`_v2`"

gsfk (changes requested):
Re-testing after katsu updates: 
- adding database Linked Field set is broken 
- if you try to edit database info more than once, the information is often stale and behaves weirdly
- editing sometimes modifies the wrong dataset (I had dataset "A" and dataset "B".... editing the name of A changed B instead and left A untouched)

gsfk (changes requested):
Form data still looks stale, if I click through multiple datasets in turn, I get information for the previous dataset I clicked rather than the current one. 

Also having trouble with a dataset creation features, lots of stuff not related to provenance is not working. Will this magically fix if we stop naming stuff "v2"?
- can't ingest data into a new dataset (katsu says `error encountered while ingesting: dataset does not exist` ). But ingestion _does_ work for the v2 datasets that were created by the migration process. 
- data summaries are broken (the modal when you click on "Clinical Data", "Experiments" or "Variants".. values are all zero or "N/A")

SanjeevLakhwani:
> does it do anything with the GeoJSON or just store it as a string?

Nothing for now, it is just a possible input for a field in provenance

src/components/datasets/DatasetForm/tabs/ContactsTab.tsx
```diff
@@ -0,0 +1,66 @@
...
+      size="small"
+      style={{ marginBottom: 8 }}
+    >
+      <PersonOrOrganizationFields namePrefix={["primary_contact"]} form={form} />
+    </Card>
+
+    <Card
+      title={
+        <span>
+          Stakeholders <RequiredMark />
```
davidlougheed: stakeholders shouldn't be required - just primary contact

src/components/datasets/DatasetForm/tabs/CoreInfoTab.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <Form.Item label="Release date" name="release_date">
+        <DatePicker style={{ width: "100%" }} />
+      </Form.Item>
+
+      <Form.Item label="Last modified" name="last_modified">
+        <DatePicker style={{ width: "100%" }} />
+      </Form.Item>
+
+      <Form.Item label="Discovery config" name="discovery">
+        <DiscoveryFileSelect />
```
davidlougheed: what was wrong with the previously-used `DropBoxJsonSelect`? adding a whole new component with roughly redundant behaviour is a bit too much ai-generated duplication for me

src/components/datasets/DatasetForm/tabs/PcglInfoTab.tsx
```diff
@@ -0,0 +1,50 @@
...
+import { MinusCircleOutlined, PlusOutlined } from "@ant-design/icons";
+
+const { Option } = Select;
+const { Text } = Typography;
+
+const PcglInfoTab = () => (
+  <>
+    <Card title="Study" size="small" style={{ marginBottom: 8 }}>
+      <Form.Item label="Study status" name="study_status">
+        <Select placeholder="Select status" allowClear>
```
davidlougheed: use `options` prop, not `Option` jsx def. same goes for all selects in the PR

src/modules/datasets/actions.js
```diff
@@ -13,7 +13,7 @@ export const FETCH_DATASET_RESOURCES = createNetworkActionTypes("FETCH_DATASET_R
 const fetchDatasetDataTypesSummary = networkAction((serviceInfo, datasetID) => ({
   types: FETCH_DATASET_DATA_TYPES,
   params: { serviceInfo, datasetID },
-  url: `${serviceInfo.url}/datasets/${datasetID}/data-types`,
+  url: `${serviceInfo.url}/${(serviceInfo.bento?.serviceKind ?? serviceInfo.type?.artifact) === "metadata" ? "datasets_v2" : "datasets"}/${datasetID}/data-types`,
```
davidlougheed: put TODOs for these janky ternaries to remove them in next version (if theyre still needed?)

src/components/datasets/DatasetFormModal.js
```diff
@@ -12,8 +14,10 @@ import { useProjects } from "@/modules/metadata/hooks";
 import { datasetPropTypesShape, projectPropTypesShape, propTypesFormMode } from "@/propTypes";
 import { nop } from "@/utils/misc";
 import { useAppDispatch } from "@/store";
+import { saveDraft, loadDraft, clearDraft, deserializeFormValues } from "@/utils/datasetDraftUtils";
```
davidlougheed, line 16: group with other util imports

src/components/datasets/DatasetForm/tabs/LinksMediaTab.tsx
```diff
@@ -0,0 +1,75 @@
...
+      <Form.List name="logos">
+        {(fields, { add, remove }) => (
+          <>
+            {fields.map(({ key, name }) => (
+              <Card key={key} size="small" style={{ marginBottom: 8 }}>
+                <Form.Item label="URL" name={[name, "url"]} rules={[{ required: true, type: "url" }]}>
+                  <Input placeholder="https://..." />
+                </Form.Item>
+                <Form.Item label="Theme" name={[name, "theme"]}>
+                  <Select>
```
davidlougheed: use options prop not jsx def

src/components/datasets/DatasetForm/fields/PublicationVenueFields.tsx
```diff
@@ -0,0 +1,60 @@
...
+  absoluteNamePrefix?: (string | number)[];
+}) => {
+  const absPrefix = absoluteNamePrefix ?? namePrefix;
+  return (
+    <Card size="small" title="Venue" style={{ marginBottom: 8 }}>
+      <Form.Item label="Venue name" name={[...namePrefix, "name"]} rules={[{ required: true, min: 1 }]}>
+        <Input />
+      </Form.Item>
+      <Form.Item label="Venue type" name={[...namePrefix, "venue_type"]} rules={[{ required: true }]}>
+        <Select placeholder="Select venue type" allowClear>
```
davidlougheed: use options prop not jsx def

src/components/datasets/DatasetForm/tabs/CoreInfoTab.tsx
```diff
@@ -0,0 +1,101 @@
...
+        <>
+          <Form.Item label="Content" name={["long_description", "content"]} rules={[{ required: true, min: 1 }]}>
+            <TextArea rows={6} placeholder="Extended description..." />
+          </Form.Item>
+          <Form.Item
+            label="Content type"
+            name={["long_description", "content_type"]}
+            rules={[{ required: true, message: "Content type is required" }]}
+          >
+            <Select placeholder="Select content type">
```
davidlougheed: use options prop not jsx def

src/components/datasets/DatasetForm/fields/PersonOrOrganizationFields.tsx
```diff
@@ -0,0 +1,116 @@
...
+}) => {
+  const absPrefix = absoluteNamePrefix ?? namePrefix;
+  // eslint-disable-next-line react-hooks/exhaustive-deps
+  const typePath = useMemo(() => [...absPrefix, "type"], [JSON.stringify(absPrefix)]);
+  const typeValue = Form.useWatch(typePath, form) ?? "person";
+
+  return (
+    <Card size="small" style={{ marginBottom: 8 }}>
+      <Form.Item label="Type" name={[...namePrefix, "type"]}>
+        <Radio.Group
```
davidlougheed, line 91: use options prop; see https://ant.design/components/radio

davidlougheed (changes requested):

src/components/datasets/DatasetForm/tabs/CoreInfoTab.tsx
SanjeevLakhwani: mb I should have noticed that there was this class.

src/components/datasets/DatasetForm/tabs/ClassificationTab.tsx
```diff
@@ -0,0 +1,189 @@
...
+      </Form.List>
+    </Card>
+
+    <Card title="Taxonomy" size="small" style={{ marginBottom: 8 }}>
+      <Form.List name="taxa">
+        {(fields, { add, remove }) => (
+          <>
+            {fields.map(({ key, name }) => (
+              <Card key={key} size="small" style={{ marginBottom: 8 }}>
+                <OntologyFields listName="taxa" name={name} />
```
davidlougheed: ideally, we should have a rapid option for `NCBITaxon:9606` (*Homo sapiens*) and the placeholder should match the taxonomy field.

src/components/datasets/DatasetForm/tabs/ClassificationTab.tsx
```diff
@@ -0,0 +1,189 @@
...
+            ))}
+            <Button type="dashed" onClick={() => add({ type: "string" })} icon={<PlusOutlined />}>
+              Add taxonomy entry
+            </Button>
+          </>
+        )}
+      </Form.List>
+    </Card>
+
+    <Card title="Ontology Resources" size="small" style={{ marginBottom: 8 }}>
```
davidlougheed: ideally we should pre-fill this a bit but since that'll be complex that can come in the future.

src/components/datasets/DatasetForm/tabs/ClassificationTab.tsx
```diff
@@ -0,0 +1,189 @@
...
+    </Card>
+
+    <LicenseSection />
+
+    <Card title="Spatial Coverage" size="small">
+      <Text type="secondary" style={{ display: "block", marginBottom: 8 }}>
+        Provide a place name string or a GeoJSON Feature (as JSON).
+      </Text>
+      <Form.Item label="Coverage (string)" name="spatial_coverage">
+        <Input placeholder="e.g. Canada, North America" />
```
davidlougheed: does this actually support geojson at all?

gsfk (commented):
- shouldn't there be a field for data access codes?
- geoJson serializes weirdly: 

Image

src/components/datasets/DatasetForm/tabs/ClassificationTab.tsx
SanjeevLakhwani: you put it as a json string and it would detect it.

gsfk (approved):
Preliminary approval since I'm out of issues for this latest iteration. No comments on code.

It's now possible to have ontology resource mismatches between the dataset info, and the data itself, by each having different versions, but:
- that's now easy to edit for the dataset info
- it's not clear that it even matters

For actually using the ontologies in phenopacket view in bento_public, it seems odd that we don't simply use the resources that are in the phenopacket itself. But this is a separate issue.

davidlougheed (approved):
lgtm

PR #541, opened 2026-04-08
feat(authz): support editing grants via PUT /grants/:id

SanjeevLakhwani:
## Summary

- Adds an Edit button to the grants table that opens a pre-filled edit modal
- Implements `saveGrant` action using `PUT /grants/:id` with optimistic UI update in the reducer
- Fixes a bug in `PermissionsInput` where permissions were wiped on first modal open because the `currentResource` effect ran before the permissions list loaded, filtering all IDs to an empty array

## Changes

- **`GrantsTabContent.js`**: Added `GrantEditModal` component, edit state, and wired up the Edit button (previously commented out)
- **`actions.js`**: Added `SAVE_GRANT` action types and `saveGrant` action creator
- **`reducers.js`**: Added `isSaving` state and `SAVE_GRANT` reducer cases with optimistic update; on error, invalidates grants to trigger a re-fetch
- **`GrantForm.tsx`**: Accepts an `initialValues` prop passed as form-level `initialValues` (higher priority than field-level) to correctly pre-fill the edit form on first open; fixed `PermissionsInput` resource-change effect to guard against empty `permissionsByID`

### DIFF
--- a/src/components/manager/access/GrantsTabContent.js
+++ b/src/components/manager/access/GrantsTabContent.js
@@ -113,7 +113,7 @@ const GrantsTabContent = () => {
 
   const isFetchingAllServices = useServices().isFetchingAll;
 
-  const { data: grants, isFetching: isFetchingGrants } = useGrants();
+  const { data: grants, itemsByID: grantsById, isFetching: isFetchingGrants } = useGrants();
 
   const {
     isFetching: isFetchingPermissions,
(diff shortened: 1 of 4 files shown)


src/components/manager/access/GrantsTabContent.js
```diff
@@ -16,6 +16,48 @@ import GrantForm from "./GrantForm";
...
+  const onOk = useCallback(() => {
+    setLoading(true);
+    form
+      .validateFields()
+      .then(async (values) => {
+        await dispatch(saveGrant({ ...values, id: grant.id }));
+        closeModal();
+      })
+      .catch((err) => {
+        console.error(err);
```
davidlougheed, line 33: give some kind of user feedback beyond this via antd `message`

davidlougheed (changes requested):
first pass

davidlougheed (changes requested):
form doesn't clear itself properly after clicking on different grants, resulting in overwriting grant contents which is very undesirable!

davidlougheed (approved):
lgtm; once gordon approves it i think we're good to merge.

gsfk (approved):

PR #552, opened 2026-05-29
reference genomes tab:fix link to gff3 tbi file

gsfk:
Fix link to gff3 tbi index file (currently just repeats the link to the gff3 file).


davidlougheed (approved):
oops, thanks!

PR #559, opened 2026-06-25
Don't allow editing of grant resource or subject

gsfk:
Simplest possible fix for Redmine 2814: when editing grants, don't allow changes to "subject" or "resource". 

- adds an `isEditing` parameter to some grant form components... this could have been avoided by defining `isEditing=!!initialValues` but that seems likely to break something in the future
- removed some comments saying that grants are not editable
- leaves open for the moment what to do about resources that no longer exist, this is arguably a separate issue (the UI already makes it clear that they don't exist, currently it's just up to the user to remove them).


davidlougheed (changes requested):
changing subject group should also be disabled given the scope of this PR

davidlougheed (approved):
lgtm

